PR #3, opened 2017-01-24
allow appending unregistered classes , models if provided as strings. etc

jheld:
allow just appending unregistered classes
get the class for models if provided as strings.
crud event keep user id if deleted
store datetime with timezone info


soynatan:
Shortly I'll update readme file to explain these settings.

soynatan:
Setting names have been changed and added to readme file. Thank you for these contributions!

PR #4, opened 2017-01-25
allow crud difference callbacks to determine of CRUDEvent should be created

jheld:
allow crud difference callbacks; if any false do not create crud
use dispatch_uid on signal connections so they are guaranteed just one run per that module import

### DIFF
--- a/easyaudit/settings.py
+++ b/easyaudit/settings.py
@@ -1,26 +1,25 @@
+from importlib import import_module
 
+from django.apps.registry import apps
 from django.conf import settings
-
-
-# django unregistered classes
-from django.db.migrations import Migration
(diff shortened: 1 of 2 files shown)


soynatan:
Thank you for these contributions. They have been merged into master. Keep them coming! :)

The only thing I changed back are the names of the settings. I wasn't sure about the word "append" because it only makes sense if you know how the app works. So I went for these names:

`DJANGO_EASY_AUDIT_UNREGISTERED_CLASSES_DEFAULT` for default unregistered classes.

`DJANGO_EASY_AUDIT_UNREGISTERED_CLASSES_EXTRA` for user defined (extra) unregistered classes.

I think with these changes I'll release v0.8. Kudos!

PR #13, opened 2017-03-22
middleware init with optional kwarg.

jheld:
middleware init with optional kwarg.

I was getting a startup error because the `__init__` apparently wants no arguments. The mixin class's `__init__` had it as a default variable, so I did the same (I could have just not subclassed the init of course).


soynatan:
Great! Thank you, it's been merged.

PR #20, opened 2017-07-05
Log `m2m_changed` signal

KatrinaE:
This PR updates django-easy-audit to log m2m-changed signals. I'm still fixing the following two issues:

1) Each M2M update is recorded twice, because M2M changes send two signals (`pre_add`/`post_add` for additions and `pre_remove`/`post_remove` for deletions).

   * I am going to filter out half of the signals. I would prefer to keep the pre-save actions (`pre_add` and `pre_remove`), but to be consistent with easyaudit's other signals, I am going to keep the post-save ones (`post_add` and `post_remove`).

2) Reverse M2M updates do not include the M2M change in the audit log.

   * Reverse M2M updates (e.g. if the `Business` model has a ManyToManyField connected to `User`, a reverse add would be `some_user.business_set.add(some_business)`, as opposed to `some_business.users.add(some_user)`) do not include the M2M field, because the M2M field is not actually present on that model. I'm not sure how to resolve this; need to think about it more.

### DIFF
--- a/easyaudit/models.py
+++ b/easyaudit/models.py
@@ -8,10 +8,15 @@ class CRUDEvent(models.Model):
     CREATE = 1
     UPDATE = 2
     DELETE = 3
+    M2M_CHANGE = 4
+    M2M_CHANGE_REV = 5
+
     TYPES = (
         (CREATE, 'Create'),
         (UPDATE, 'Update'),
         (DELETE, 'Delete'),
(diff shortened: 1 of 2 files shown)


soynatan:
This looks interesting! I completely ignored that signal. I will merge your code to m2m-changed branch, and when it is ready we can merge it to master. Thank you :)

PR #28, opened 2017-08-23
Removed redundant save() calls (create() includes implicit save())

rossettistone:
Hi @soynatan! Thanks for your work on this library. We were on the hunt for a django auditing lib, and yours was just the thing!

After integrating, we noticed when running our tests that there were more audit queries than we expected. A little digging led us to find two `save()` calls in `signals.py` that look redundant with the `create()` calls above them. I'm a bit new to django development, so I checked the relevant docs to be sure:
https://docs.djangoproject.com/en/1.11/ref/models/querysets/#create

This PR just removes the redundant `save()` calls, which should be a nice performance boost for those (like us!) using the library heavily in their apps. 

I ran the tests after making this change, and also did some live tests in our app (not referenced in this PR, naturally), and things still seem to be working.

Let me know if you have any thoughts on this PR. Happy to make changes if helpful!

### DIFF
--- a/easyaudit/signals.py
+++ b/easyaudit/signals.py
@@ -147,8 +147,6 @@ def m2m_changed(sender, instance, action, reverse, model, pk_set, using, **kwarg
             datetime=timezone.now(),
             user_pk_as_string=str(user.pk) if user else user
         )
-
-        crud_event.save()
     except Exception:
         logger.exception('easy audit had an m2m-changed exception.')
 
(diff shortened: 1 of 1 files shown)


soynatan:
You are absolutely right. I think that in previous code versions the object was created as an instance and then saved. At some point that was replaced with `create()`, but the `save()` was overlooked.

I'm merging your PR, thank you so much!

rossettistone:
Wonderful! Glad to be able to help!

PR #34, opened 2017-10-20
Make it run with Django 2.0

flesser:
`on_delete` on ForeignKeys is required since Django 2.0, see https://docs.djangoproject.com/en/1.11/ref/models/fields/#foreignkey


soynatan:
Hello, thank you for your contribution. I only have one doubt: are we sure it should be `models.CASCADE` and not `models.SET_NULL` (with `null=True`)?

flesser:
`models.CASCADE` is the current behaviour:
> `on_delete` will become a required argument in Django 2.0.
> **In older versions it defaults to `CASCADE`**.

https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.ForeignKey.on_delete

soynatan:
Yes, I just thought maybe it was the moment to rethink if that behaviour was the best choice. I'll be merging in a minute. Thanks.

PR #39, opened 2017-11-10
Bugfix: check if HTTP_COOKIE is truthy, and use django's version.

jheld:
Unfortunately, I ran into issues during my project's unit tests. This seems to work. Please check on your end, too.


soynatan:
Seems to be working fine with:

* Django 1.8 with Python 2

* Django 1.8 with Python 3

* Django 1.11 with Python 2

* Django 1.11 with Python 3

PR #40, opened 2017-11-11
Improved admin: Added filters and hyperlinks to admin; "Purge" action added to let superuser remove all rows from EasyAudit tables.

morlandi:
Just standard Django admin staff added for easier search and navigation of collected audit data,
plus "purge" action added to let superuser remove all rows from EasyAudit tables

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,4 @@
 include LICENSE
 include README.rst
+recursive-include easyaudit/static *
 recursive-include easyaudit/templates *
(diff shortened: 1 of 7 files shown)


soynatan:
Wow, this is a great improvement to the admin! Before merging, I have a question: is there a reason why you are showing `object_json_repr_prettified` AND `object_json_repr` in the detail template? Seems to me that `object_json_repr_prettified` is enough.

Thank you for all the work!

morlandi:
Hi, that was just in case someone would like to copy and paste the Json serialization in a compact form ... however, I do agree it's fully redundant; you might very well get rid of it

morlandi:
Thank you @soynatan

soynatan:
Thank _you_!

PR #43, opened 2018-01-05
added missing migration

atakanarikan:
I realized some of the changes were not reflected with migrations, therefore running `makemigrations` in my app would yield in a new migration.
Added the missing migration to the library


rcmontoy:
Thank you for creating this PR, @atakanarikan. I would love to see this merged 👍

PR #51, opened 2018-03-01
Feature/delta update

### LINKED ISSUE #24: Delta and full update JSON on CRUDEvent
My team is considering having a delta field of the changes, as opposed to the full JSON object on `post_save`.

I haven't started my implementation yet, but I was curious what you thought of the following extension:

- In the signal code, let the output of the callback/hook code (as we know, it's optional), return False, or True, or some sort of enum-like value. In this case it could be FULL, DELTA. That could be a little confusing, but maybe not so bad. Or, have it be a different kind of hook -- different from the one currently implemented in the code.

- Allow saving both the FULL and DELTA on the CRUDEvent, so the project can determine if they want both or one or the other, always, or programmatically. This is in the same vein as above. It unfortunately would potentially duplicate the JSON in the worst case, if both are used on a given object operation.

I want to keep it as simple as possible, but still flexible for the consumer's needs. Are you interested in such a change?

soynatan:
Closes #24 :)


easyaudit/signals/model_signals.py
```diff
@@ -71,7 +80,54 @@ def post_save(sender, instance, created, raw, using, update_fields, **kwargs):
                                     for callback in CRUD_DIFFERENCE_CALLBACKS if callable(callback))
 
             # create crud event only if all callbacks returned True
-            if create_crud_event:
+            if create_crud_event and not created:
```
jheld: Ah, I see what you've done.

On pre-save, you add in the changed_fields, if we're an edit, not a creation.

And on post-save, you save a new crud event if the object is newly created (and thus no changed_fields).

Curious, why could all of this logic not be done in the post_save handler? As in, why make the two cases completely separate like this, handler wise?

PR #58, opened 2018-04-05
Added code to check for loading fixtures

manishgupta24:
Return from signal if a fixture is loading. 

https://docs.djangoproject.com/en/2.0/ref/signals/#pre-save
https://docs.djangoproject.com/en/2.0/ref/signals/#post-save

Check raw param.


jheld:
@manishgupta24 thank you for this PR! We're looking to plan a release in the short term so it would be good to get this in.

Can the other signals use this, too?

PR #66, opened 2018-07-12
Made the admin LogEntry model optional for apps that don't use the Django Admin

robmagee:
Ignore my prior pull request- had the new setting set backwards.

The LogEntry import was causing an error in my application since I'm not using the django admin app. I made an extra setting DJANGO_EASY_AUDIT_USE_LOG_ENTRY which by default is True and keeps the expected/current behavior. When set to false, it will not import the LogEntry model from the admin
app and not cause the error.

### DIFF
--- a/easyaudit/settings.py
+++ b/easyaudit/settings.py
@@ -2,7 +2,6 @@ from importlib import import_module
 
 from django.apps.registry import apps
 from django.conf import settings
-from django.contrib.admin.models import LogEntry
 from django.contrib.auth.models import Permission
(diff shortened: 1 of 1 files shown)


soynatan:
Hey, I took your suggestions but without the need to add another config option. Now easy-audit will check if you are using the django admin app in your project, and then import LogEntry or not. See https://github.com/soynatan/django-easy-audit/commit/858403be7cd9718e9d331747ed6eac538c0e65e6. Thank you!

Issue #74, opened 2018-12-11
pre_save signal fails serialization for new objects


andresdevsar:
When the signal tries to serialize the django User model it fails because the object doesn't have a PK yet
for many to many relation

`object_json_repr = serializers.serialize("json", [instance])`

https://github.com/soynatan/django-easy-audit/blob/master/easyaudit/signals/model_signals.py#L56

jheld:
#80 should fix this (at least the bug). We will want to fix it in an algorithm fix, but at least we can guard against the exception.

Issue #71, opened 2018-09-03
removal of db_constraint on User


jheld:
Hello!

My project was/is looking to separate our easyaudit data to a separate database.

What would your thoughts be on removing the db_constraint on the models that reference a User?

This would mean that if the User doesn't exist that an exception would be raised when trying to access it.

However under normal flow it would be fine for most applications.

I'm not sure if there are other side effects from doing this, but I wanted your thoughts. I'm also not sure that this could be a project-level setting (as in it may have to be hard-coded).

jheld:
My team will be looking into this in the short term. We're going to find out how feasible it is and what changes are needed in this project to accommodate it.

soynatan:
Hello, sorry, I haven't been able to work on easy audit. I still need to close another project before I can get my hands on this. If you and your team will be working on this, that would be great to know the implications. Thank you.

soynatan:
Closing as it has been implemented in #80. Thank you for your work.

PR #75, opened 2019-01-03
Changed version to 1.1

funcho:
Version number changed to 1.1 for new release with latest bugfixes (https://github.com/soynatan/django-easy-audit/pull/58 and https://github.com/soynatan/django-easy-audit/pull/61)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ os.chdir(os.path.normpath(os.path.join(os.path.abspath(__file__), os.pardir)))
 
 setup(
     name='django-easy-audit',
-    version='1.0',
+    version='1.1',
     packages=find_packages(),
     include_package_data=True,
     license='GPL3',


jheld:
cool!

We were able to successfully implement the user & content type db constraint removal (#71 ). If merged, perhaps it could go in 1.1 or 1.2?

#74 also slightly "fixed" it by ignoring the error when it occurs. We can get around this instead by using the previous CRUDEvent on the post_save, to emulate the change set. Not in this scope, but just stating an alternative.

funcho:
We could create release 1.1 for now and when those changes are merged then version 1.2 would be created.

funcho:
@jheld Is it posible to do that?

jheld:
Yes, I think so.

I'll try to get my PR ready early next week.

I think I can merge this PR. If so, I'll check what's coming in since the last release and then pull the trigger. @soynatan we will also need a PyPI release.

jheld:
Sorry that I haven't yet merged or made another PR, yet.

I'm also a bit hesitant to do the merge without having a PyPI release queued up shortly after. Everyone would have to put to github instead...not a great solution. Or make a custom `whl`, also not great.

PR #77, opened 2019-01-23
Disable pre-save and post-save signals for loaddata

graceshaw:


soynatan:
This looks good to me. I will merge it later today, unless there's something to add. Thank you.

soynatan:
Hey, @graceshaw, pre and post save signals should already be bypassed when loading fixtured. Check this commit: https://github.com/soynatan/django-easy-audit/commit/9b67a36081c1cf2bd12d294315193f060b097319#diff-f189a107dab08d3d973ab89d7198619c

Isn't it working for you? Or maybe you have an older version?

graceshaw:
@soynatan It wasn't working for us (with the latest version)

jheld:
Right. I wondered the same. Can you tell us what the value for `raw` was? I want to be sure that there isn't something nefarious going on.

jheld:
Closing for now, please re-open if you'd like to discuss using this flow instead of the current way we skip load-data. Or if it's still not working!

PR #80, opened 2019-02-21
split user db constraint, content type retrieve and create fix

jheld:
This implements #71 .

This is currently in production at my company and we've had essentially no issues. There may need to be some adjustments/fixes on the crud deletion path, but am not sure if it's from this code change or not.

For users that don't need to split their database (removing the constraint), there is just a migration that will run, but otherwise there will be no effective change.

I have added some extra detail into the README, though admittedly I didn't go into much detail. That can be fixed before merge if needed.


easyaudit/migrations/0009_auto_20180314_2225.py
```diff
@@ -20,6 +20,6 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='requestevent',
             name='url',
-            field=models.TextField(db_index=True),
+            field=models.TextField(db_index=False),
```
jheld: this migration actually failed for us. So I'm hoping that users whom have not yet run the migration won't be affected by it, due to this change.

Any users that want to remove the index will have to do it by hand, via raw SQL.

easyaudit/migrations/0011_auto_20181101_1339.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('easyaudit', '0010_repr_text'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='requestevent',
+            name='url',
+            field=models.TextField(),
```
jheld: Actually I think this migration actually explicitly undoes the previous ones index. So that way the project is consistent.

soynatan:
I will be reviewing this this afternoon.

easyaudit/migrations/0009_auto_20180314_2225.py
dopatraman: Are you using MySQL? If so, keeping the index, changing the column type to `CharField`, and adding a `max_length` attribute did the trick for me.

easyaudit/migrations/0011_auto_20181101_1339.py
dopatraman: Awesome. Any objections to making this column a `CharField`? If it doesn't create problems for you, it'd prevent conflicts with a PR I'm currently working on: https://github.com/soynatan/django-easy-audit/pull/82/files

soynatan: I can't recall why we changed the URL to be a `TextField`. But I believe we should make it a `CharField` again. @jheld what do you think? Am I missing something? Or maybe we should leave it a `TextField` and make this change afterwards.

easyaudit/signals/request_signals.py
```diff
@@ -50,7 +50,7 @@ def request_started_handler(sender, environ, **kwargs):
         url=environ['PATH_INFO'],
         method=environ['REQUEST_METHOD'],
         query_string=environ['QUERY_STRING'],
-        user=user,
+        user_id=getattr(user, 'id', None),
```
soynatan: Is there a reason why you changed `user` to `user_id` (and also, `content_type` to `content_type_id`, etc.) when creating Event objects?

easyaudit/migrations/0011_auto_20181101_1339.py
dopatraman: @soynatan If you look at my PR #82 , I change the field to a URL.

easyaudit/migrations/0009_auto_20180314_2225.py
jheld: Yes, using MySQL. Interesting. I could see us wanting to keep as a TextField though, so there is no information loss.

On a separate PR I had suggested a url field and params field.

I don't have the luxury at this time to update the field on the fly (we are in production with this code change already). But what do you think about splitting the fields up?

easyaudit/migrations/0011_auto_20181101_1339.py
jheld: I agree charfield would be best. From the other comment, I am more inclined to splitting up url and query params to separate fields at a later date, and change the type then.

easyaudit/signals/request_signals.py
jheld: Yes. Mostly for being explicit because we only have IDs technically supported on this code change. So I do think it would fail (on my project, since the db has been split).

jheld:
@dopatraman yes, this will let it be split apart. It's been working quite well for us (over 20 million CRUDEvent objects on a separate db). Some of the queries needee to be changed (use id look ups), and our subclassed admin, too. But really it wasn't bad to migrate.

However, your project will have to manage the migrations to support having multiple databases. Let me know if you would like assistance with that.

jheld:
@soynatan how are we looking? I'm happy to wait more if you're still reviewing.

@dopatraman same goes for you.

It looked like there was general happiness, but wanted to be sure.

easyaudit/migrations/0011_auto_20181101_1339.py
soynatan: > I am more inclined to splitting up url and query params to separate fields at a later date

Maybe I'm missing something (I've been away for some time) but, we already have url and query params on separate fields.

soynatan:
> @soynatan how are we looking? I'm happy to wait more if you're still reviewing.

It all looks ok to me! We can discuss changing to a `Char` or `URL` field on another pull or issue. I'll merge it right away.

jheld:
@soynatan oh wow...I did not see that. Then yeah, I did part of that work for nothing. Let's try to fix soon.

max-wittig:
@soynatan Can we have a new release with this bug fixed?

PR #82, opened 2019-03-22
Compatibility with Django 2.x 

dopatraman:
# Goal
The goal of this PR is to allow compatibility between easy_audit and Django v 2.x.

# Implementation Details
Django 2 signal receivers do not accept an `environ` parameter anymore. Instead, this information is taken from the request scope.

# Discussion
Also, I've updated the URL column to use a Charfield, so that the library accommodates MySQL users.

### DIFF
--- a/easyaudit/migrations/0009_auto_20180314_2225.py
+++ b/easyaudit/migrations/0009_auto_20180314_2225.py
@@ -20,6 +20,6 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='requestevent',
             name='url',
-            field=models.TextField(db_index=True),
+            field=models.CharField(db_index=True),
         ),
     ]
--- a/easyaudit/migrations/0009_auto_20180314_2225.py
+++ b/easyaudit/migrations/0009_auto_20180314_2225.py
@@ -20,6 +20,6 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='requestevent',
             name='url',
-            field=models.CharField(db_index=True),
+            field=models.CharField(db_index=True, max_length=255),
         ),
     ]
--- a/easyaudit/models.py
+++ b/easyaudit/models.py
@@ -70,7 +70,7 @@ class LoginEvent(models.Model):
 
 
 class RequestEvent(models.Model):
-    url = models.TextField(null=False, db_index=True)
+    url = models.CharField(null=False, db_index=True)
     method = models.CharField(max_length=20, null=False, db_index=True)
     query_string = models.TextField(null=True)
     user = models.ForeignKey(settings.AUTH_USER_MODEL, null=True, blank=True,
--- a/easyaudit/models.py
+++ b/easyaudit/models.py
@@ -70,7 +70,7 @@ class LoginEvent(models.Model):
 
 
 class RequestEvent(models.Model):
-    url = models.CharField(null=False, db_index=True)
+    url = models.CharField(null=False, db_index=True, max_length=255)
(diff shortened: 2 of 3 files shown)


jheld:
Thank you! I have an open PR which does some migration things (fixes and features). I think that needs to go in first (to keep migration dependencies happy), since we are in production with it.

But a fair bit of what's in here definitely will want to get in, too.

I'll check back on my PR and see if it's in a good state. #80

dopatraman:
@soynatan thoughts on merging?

jheld:
Can we split out the environ issue from the migration piece?

I really want to keep the migration PRs in order if possible.

I can see about merging my PR (tomorrow?) and then let's talk about this one.

soynatan:
> Can we split out the environ issue from the migration piece?

Agree.

> @soynatan thoughts on merging?

Let's wait till #80 is merged. Then we can work on your PR, if it's OK with you.

ptr0x01:
Any update on 2.0 support?

jheld:
Absolutely! My company is on django 2.1 with master code. The one caveat is that requestevent may be broken on 2.1. 2.0 though i think should be fine. The 2.1 breakage has an open PR.

ptr0x01:
> Absolutely! My company is on django 2.1 with master code. Django 2.0 should work out of the box.

Thanks a lot, we are on 2.2 and ran into this same issue with `request_started_handler() missing 1 required positional argument: 'environ'` which I beleive was related to this pull request.

mrgoos:
Is this planned to be merged?

jheld:
Some conflicts here and still open questions regarding 2.2 and under support. We might have to open a new branch to try based off of these changes. We're going to get more projects needing this to support 2.2 so the head is on. If I make one I'll update here regarding it.

jheld:
@ptr0x01 I'm running this locally with django 2.2 and do not have this issue.

I will see if I can dig a little further, but if I cannot reproduce this issue, I think I'll update 1.1.1rc1/1.1.1 to include django 2.2 support officially. We're at 1.1.1rc1 but if 2.2 works then I'll make a setup.py adjustment on the commit before the rc removal.

jheld:
@mrgoos have you also had this issue or were you waiting for this before going to django 2.2?

jheld:
I need to apologize! The issue that was opened where we didn't have the `environ` param was back from an old version of easyaudit that probably didn't yet support django 1.8 (when that param was added). The commit where we added it was on 10/18/17. Unsure exactly when it was merged, but point is, we've already fixed it :)

I'm not sure about the rest of the code changes. Scouring through the docs (1.8 and up to 2.2), https://docs.djangoproject.com/en/2.2/ref/signals/#django.core.signals.request_started the arguments to the signal have remain unchanged. This PR looks like it would break easyaudit as is. Definitely interested to hear more if there is something I'm missing/new PR with specific changes.

I will make a 1.1.1rc2 which includes django 2.2 support in the setup.py. Shortly after that, if nothing shows up, I'll cut 1.1.1 explicitly. 1.1.2 will likely drop some versions of django support/and or python support, to keep up with just the supported versions of django.

Closing this PR for now. Please re-open if needed!

jheld:
OK, we do not technically to update setup.py. I have included a new commit on master so that the new release will have updated trove classifiers, but the explicit required versions in the setup.py restrictions does not exist :) That should be changed! But we're safe for now. We should have a separate PR which looks to include real version restrictions, but since so far there hasn't been any new code change for 1.1.1 (aside from a migration db fix), I do not see any reason to update the setup.py for real.

Issue #69, opened 2018-08-07
KeyError: '_auth_user_id'


vcostas:
"/usr/local/lib/python3.5/dist-packages/easyaudit/signals/request_signals.py", line 40, in request_started_handler
    user_id = session.get_decoded()['_auth_user_id']
KeyError: '_auth_user_id'

Please merge the code. 
Fix in fork 'https://github.com/vcostas/django-easy-audit'

mugumedavid:
@vcostas I've made a PR from my fork that resolves this. I experienced this bug as well and resolved it that way.

PR #85, opened 2019-04-01
Defensive check for user_id in session data

mugumedavid:
#69 issue resolution


jheld:
hi! seems this is still an issue (obviously I'm quite delayed in reviewing this).

can you try writing up a test? Otherwise agreed, pretty straightforward change.

Issue #92, opened 2019-06-05
Add support for custom serializers


jdkizer9:
Some of my models that I would like to audit contain sensitive information that'd prefer not be saved. Ideally, I would have the option of fields to ignore on a per model basis when performing json serialization. I think that this could be accomplished with something as simple as a callback that post serialization that would allow for modifying the json. Additionally, for changes to sensitive fields, it would be nice to see that a change to that field was made without showing the info itself.

jdkizer9:
Opened #93 to address this. Closing

PR #93, opened 2019-06-05
Feature/custom object json serializer

jdkizer9:
Fixes issue #92


jdkizer9:
Some of my models that I would like to audit contain sensitive information that'd prefer not be saved. Ideally, I would have the option of fields to ignore on a per model basis when performing json serialization. I think that this could be accomplished with something as simple as a callback that post serialization that would allow for modifying the json. Additionally, for changes to sensitive fields, it would be nice to see that a change to that field was made without showing the info itself.

jdkizer9:
Below is an example usage.

Create the following methods:

```
def excluded_field_names_for_instance(instance):
    if instance._meta.label == 'auth.User':
        return ['password']
    if instance._meta.label == 'app.Token':
        return ['token']
    return []

def custom_easy_audit_serializer(instance):
    excluded_fields = excluded_field_names_for_instance(instance)
    field_names = [field.name for field in instance._meta.fields if field.name not in excluded_fields]
    return serializers.serialize("json", [instance], fields=field_names)

def easy_audit_model_delta_callback(old_instance, new_instance, delta):
    excluded_fields = excluded_field_names_for_instance(new_instance)
    for field_name in excluded_fields:
        if field_name in delta:
            delta[field_name] = ['excluded', 'excluded']
    return delta
```

And add the following to settings:

```
DJANGO_EASY_AUDIT_CRUD_OBJECT_JSON_REPR_SERIALIZER_OVERRIDE = custom_easy_audit_serializer
DJANGO_EASY_AUDIT_CRUD_OBJECT_MODEL_DELTA_CALLBACK = easy_audit_model_delta_callback
```

jheld:
@jdkizer9 hi. Sounds like a good use case.

I do wonder though if for instance the serializer had an exception should it fall back to the default? At scale you may want that ability (and normalize the data after the fix has been issued on the client code).

jheld:
@jdkizer9 can you resolve the conflict and check on the comment I made?

jdkizer9:
@jheld Ok, yeah, that makes sense. Before I fix, I feel like we might also want to override serialization in [`m2m_changed`](https://github.com/jdkizer9/django-easy-audit/blob/4a9269bb847967ea41b812a8bded91f76abaf46e/easyaudit/signals/model_signals.py#L211) and [`post_delete`](https://github.com/jdkizer9/django-easy-audit/blob/4a9269bb847967ea41b812a8bded91f76abaf46e/easyaudit/signals/model_signals.py#L271). Thoughts?

jheld:
@jdkizer9 Your question re:`m2m_changed` and `post_delete` I think I agree. Is there any reason not to apply it consistently?

Minor (but important overall), I think we'll need to support callables and as well as string module paths to the new settings.

In which case, I think we can add that resolver "statically" probably in `signals.py` after the import.

mmoravcik:
@jdkizer9 @jheld what is the conclusion here? Ability to hide sensitive fields seems like a crucial one for audit system

fbarcia-sirt:
Good morning, at the end, this functionality is implemented? I don't know how to hide some fileds from my models, for example, password filed (it is encrypted but I don't want to show this field).


Thanks!

Issue #97, opened 2019-07-17
Literal string 'null' stored in changed_field


jameskirsop:
I'm seeing a literal string of `'null'` being stored in the database against some Update events.

If there aren't any changed fields one of the following should be occurring instead:

1. The event isn't logged as an 'Update' at all - as nothing has changed on the object
2. The changed_fields column should be left empty - `null` - instead of `'null'`.

This bug makes it tedious to `.filter()` for actual changes to a given object.

guy881:
I would opt for number 1. @soynatan @jheld, what do you think about this?

gghildyal:
Number 1 is addressed now with https://github.com/soynatan/django-easy-audit/pull/106.

jameskirsop:
Would appreciate if @gghildyal's PR could be merged into master!

jheld:
I'm happy with option 1, as well. Likely to merge #106 shortly.

If option 1 has any process issues for certain projects, then we can enable a new setting which allows doing the update and saving it as `NULL` instead. But for now, simpler seems better.

jheld:
merged and released to PyPI as 1.2.2b1.

there might be a couple other changes in this release so I'm not marking as release candidate yet. Once we've frozen 1.2.2's changeset, we'll do (probably just 1) release candidate, and then after small testing, official release.

jameskirsop:
Thanks @jheld!

PR #98, opened 2019-07-17
Fix for Issue #97

jameskirsop:
This PR changes the behaviour of 'Update' events so that no value is stored in a CRUDEvent objects `changed_fields` attribute instead of a literal 'null' string if nothing actually changed on the record.

This is a short-term / quick fix for #97.


jheld:
I wonder if this would need to be a breaking change. Technically the data is supposed to be a json string and I don't know if any projects are making that assumption full stop or are assuming try/catch or if/else logic if not.

But I agree it would be a better query pattern and more efficient for db storage to make this change.

jheld:
@jameskirsop can you share your thoughts on my above comment? That's the one thing that worries me about the change, otherwise I think it's great.

jheld:
I think this was fixed separately via #106 different, but semantically equivalent?

Issue #55, opened 2018-03-29
Update that Changed URL fields from CharField to TextField leads to issue on MS SQLServer


jdkizer9:
Running migration 0009_auto_20180314_2225 causes the following error:

```
Running migrations:
  Applying easyaudit.0009_auto_20180314_2225...Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/db/backends/utils.py", line 85, in _execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python3.6/site-packages/sql_server/pyodbc/base.py", line 520, in execute
    return self.cursor.execute(sql, params)
pyodbc.ProgrammingError: ('42000', "[42000] [Microsoft][ODBC Driver 13 for SQL Server][SQL Server]Column 'url' in table 'easyaudit_requestevent' is of a type that is invalid for use as a key column in an index. (1919) (SQLExecDirectW)")
```

imran-iq:
Same issue with mysql

```
django.db.utils.OperationalError: (1170, "BLOB/TEXT column 'url' used in key specification without a key length")
```

jheld:
@imran-iq thanks for noting this, too.

django docs regarding text (really char) field:

https://docs.djangoproject.com/en/2.0/ref/models/fields/#django.db.models.CharField.max_length


@jdkizer9 I haven't looked into this yet. Have you learned anything further?

rnegron:
The issue is the following snippet in the file `easyaudit/migrations/0009_auto_20180314_2225.py`: 

```python
migrations.AlterField(
            model_name='requestevent',
            name='url',
            field=models.TextField(db_index=True),
        ),
```
Running this migration gives the following warning: 

_"easyaudit.RequestEvent.url: (fields.W162) MySQL does not support a database index on longtext columns. HINT: An index won't be created. Silence this warning if you don't care about it."_

Possible quick fixes: 
1. Change `TextField` to `CharField` with `max_length` property and keep the `db_index`.
2. Remove the `db_index` from the `TextField`.

See: https://code.djangoproject.com/ticket/27859 and corresponding commit https://github.com/django/django/commit/538bf43458a147b7edeb7118c9f325c3f59ff6fb

Notes: Running on Django 2.1 and MySQL 5.6.10.

UmSenhorQualquer:
I got the same error with MariaDB. 
I dropped the index from the url field manually and it worked.

[...]

PR #100, opened 2019-07-24
Bugfix: RequestEvent url as charfield w/index.

jheld:
Undoes from issue #55 .

Caveat, for anyone that has run the migration, it seem safest to change the field type, and add the index.

For anyone that hasn't run that earlier migration, they should now be able to run it (just have no index), and then this new one will create the index. It's not as simple as I'd like, but I don't want to cause anyone runtime bugs during migration with SQL errors.


jheld:
@jdkizer9 would this work for you?

ahoneycutt:
@jheld Ran against backend testing ~1800 tests without incident. One thing to note is that I am not using the Request Events feature (ie `DJANGO_EASY_AUDIT_WATCH_REQUEST_EVENTS = False`)

jheld:
I haven run the custom branch locally (with request events on and on admin). Using MySQL. Migration ran. Django admin displays the rows, I can filter by user_id and by url prefix. Smoke tests are good!

jheld:
@ahoneycutt  @jdkizer9  new release! 1.1.1rc1. Will go 1.1.1 official once its running good on someone's production. I'm 3 weeks out from getting it on production, but will have it on our dev env in a week.

ahoneycutt:
Very excellent. I will see if we can target 1.1.1rc1 and give you any feedback. Thanks much!

PR #106, opened 2019-10-10
Do not create CRUDEvent on update if none of the fields has changed.

guy881:
Hello! Great project, thanks a lot!
I updated the code to not create CRUDEvent if none of the fields has changed.

### DIFF
--- a/easyaudit/signals/model_signals.py
+++ b/easyaudit/signals/model_signals.py
@@ -67,6 +67,8 @@ def pre_save(sender, instance, raw, using, update_fields, **kwargs):
             if not created:
                 old_model = sender.objects.get(pk=instance.pk)
                 delta = model_delta(old_model, instance)
+                if not delta:
+                    return False
                 changed_fields = json.dumps(delta)
                 event_type = CRUDEvent.UPDATE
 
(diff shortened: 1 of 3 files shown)


easyaudit/signals/model_signals.py
```diff
@@ -67,6 +67,8 @@ def pre_save(sender, instance, raw, using, update_fields, **kwargs):
             if not created:
                 old_model = sender.objects.get(pk=instance.pk)
                 delta = model_delta(old_model, instance)
+                if not delta:
+                    return False
```
jheld: not sure about returning `False` but seems like a minor thing for now, so it's fine.

jheld:
This appears to be causing a bug in the signal handler (the receiver gets cleared out prematurely).

Will be looking to back out the change for this release and consider putting it back in with it fixed.

jheld:
my project uses factories, so it's possible there is something happening there.
the factory does create a crud event, does another thing in the post init, but then when we go to make a real next save, there is no crud event changed. so despite the tests passing, still makes me wonder.
removal of just that `delta` `if` makes it pass.

guy881:
@jheld if you could write the test for that issue, then I could try to help resolve it. :)

jheld:
My next thought is we can allow people to use this feature but give them the option to do it, and by default do the old behavior. So if anyone wants to use it/try it out, then they can do so and until we can fully fix it, the rest of the default users (or those who explicitly don't want a behavior change), can go on like normal.

I think I'll do that within this or next week.

PR #112, opened 2019-10-25
Issue-111: Add changed_fields to kwargs when calling callbacks

gghildyal:


gghildyal:
Ah, I think https://github.com/soynatan/django-easy-audit/pull/106 is a better change and I don't need an extra callback for it.

MaehMaeh:
It is a pity that this pull requests was not realized. I would like to do the audit on column level. So only write a CRUDEvent when the monitored column was changed. Since there are no settings for this, I wanted to use the callback function. But unfortunately the changed values are not passed. Can this pull requests be applied to the master?

PR #121, opened 2019-12-18
Django 3.0 support (drops 1.11 and py2 in trove)

jheld:
Tests did pass! Really minor code changes so the 3.0 support should work out of the box.

Holding off the merge until after the 1.1.2 release is out (should be short term).


Natureshadow:
>Holding off the merge until after the 1.1.2 release is out (should be short term).

Short term is when ;)?

Easy audit is the only blocker in our Django 3.0 transition now.

jheld:
1.1.2rc2 is available on PyPI. I'll merge this now and make it 1.2rc1 because we're also dropping explicit support for 1.11 (at least in the trove, and py2). Technically we _could_ go to 2.0 with just this change but really only 1.11 _may_ be affected, and it's so old now, but really the changes in here are pretty lightweight.

jheld:
ok PyPI now has 1.2rc1 and there is a release doc with release notes under our releases page here on github.

@Natureshadow and @debdolph please test at our earliest convenience.

My project has to wait a bit longer due to another library and also a possible ORM aggregate issue, but glad to help move your projects forward!

PR #127, opened 2020-02-11
Read only events admin

Etenil:
Add the setting `DJANGO_EASY_AUDIT_READONLY_EVENTS` to make CRUD events read-only from the admin UI.

Setting to True will do the following:

- Disable purge
- Set CRUDEvent delete permission to false
- Set all fields of CRUDEvent to read-only

Note: _The details view of the CRUDEvent retains a "save" button, but no change can be made._

### DIFF
--- a/easyaudit/admin_helpers.py
+++ b/easyaudit/admin_helpers.py
@@ -54,11 +54,9 @@ class EasyAuditModelAdmin(admin.ModelAdmin):
     def get_urls(self):
         info = self.model._meta.app_label, self.model._meta.model_name
         urls = super(EasyAuditModelAdmin, self).get_urls()
-        my_urls = []
-        if not settings.READONLY_EVENTS:
-            my_urls = [
-                url(r'^purge/$', self.admin_site.admin_view(self.purge), {}, name="%s_%s_purge" % info),
-            ]
+        my_urls = [
+            url(r'^purge/$', self.admin_site.admin_view(self.purge), {}, name="%s_%s_purge" % info),
+        ]
         return my_urls + urls
 
     def purge(self, request):
@@ -73,7 +71,7 @@ class EasyAuditModelAdmin(admin.ModelAdmin):
         """
 
         if settings.READONLY_EVENTS:
-            return HttpResponse('Unauthorized', status=401)
+            raise PermissionDenied
 
         def truncate_table(model):
             if settings.TRUNCATE_TABLE_SQL_STATEMENT:
--- a/easyaudit/admin_helpers.py
+++ b/easyaudit/admin_helpers.py
@@ -31,6 +31,13 @@ def prettify_json(json_string):
 
 class EasyAuditModelAdmin(admin.ModelAdmin):
 
+    def get_readonly_fields(self, request, obj=None):
+        "Mark all fields of model as readonly if configured to do so."
+        if settings.READONLY_EVENTS:
+            return [f.name for f in self.model._meta.get_fields()]
+        else:
+            return self.readonly_fields
+
     def user_link(self, obj):
         user = get_user_model().objects.filter(id=obj.user_id).first()
         #return mark_safe(get_user_link(user))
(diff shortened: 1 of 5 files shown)


jheld:
I think the change is good overall, but I do wonder if we should be taking more advantage of subclassing/different CRUDEventAdmin registration.

Some of the guided approach is good (read only is a "simple" interface), but I'd almost rather the read only stuff be as an example in the docs, rather than us trying to please everyone and possibly getting out of date as we adjust fields.

All that said, read-only as a concept for an auditing library (even from django admin) seems like a good idea to add all fields as read only by default, instead of making it an `extend`.

I do like potentially having a setting for read only events, (delete_permission etc just taking from that).

What are your thoughts from this? Am I way off?

Etenil:
@jheld I agree that my attempt at conditionally adding read-only fields in the admin form was quite bad. I've made changes to handle the read-only fields better in the admin forms, and also to automatically pick up on all fields instead of having to maintain a list by hand.

It's cool to see that you're also interested in having this feature as it's quite core for me and the company I'm doing this for. It's a requirement for me to be able to audit what superuser do even in the event that a security breach would allow a user to elevate themselves to superuser.

I understand where you're coming from by not wanting to necessarily add this to the core library as pleasing everyone would result in bloat. From a technical standpoint it's totally doable for me to shape this in the same way as the recent custom loggers. On the other hand, I believe that many users would expect an auditing library to allow read-only event logging out of the box.

I'll defer to your judgement on this, do let me know what course of action you prefer and I'll make the relevant changes.

Etenil:
@jheld Can I have a decision on this as I'd like to be able to move on from the branch in this PR and use the main package instead?

jheld:
@Etenil thanks for this and the simplification (using the read only function).

I expect to release `b3` of the current version this week. If my project doesn't have issues/quick community feedback is positive, we will be able to release this or next week.

after that I am happy to merge!

jheld:
b3 has been released, I probably won't be cutting the official release until early next week (I'm waiting to get it merged and tested on my project), but we're still planning to get this in soon after (b3 should be fine after a minimal amount of smoke testing, and then general release!).

Etenil:
@jheld I see that the release of 1.2.2 has now been done, any idea of when this PR can be merged? Did your testing reveal any problem?

jheld:
Hey. I'm sorry I haven't come back to this.
I have a day off tomorrow so although I cannot promise I'll be able to review again and test over the next few days, I may have some time to do so.

If you don't hear back within a week, please ping back again.

easyaudit/admin_helpers.py
```diff
@@ -51,6 +58,11 @@ def user_link(self, obj):
     def has_add_permission(self, request, obj=None):
         return False
 
+    def has_delete_permission(self, request, obj=None):
+        if settings.READONLY_EVENTS:
```
jheld, line 62: Minor (non-blocking):
```suggestion
        return not settings.READONLY_EVENTS
```

easyaudit/admin_helpers.py
```diff
@@ -8,7 +8,7 @@
 except: # Django < 2.0
     from django.core.urlresolvers import reverse
 
-from django.http import HttpResponseRedirect
+from django.http import HttpResponseRedirect, HttpResponse
```
jheld, line 11: Minor (non-blocking), but it looks like we don't need `HttpResponse` imported under the current change.

easyaudit/tests/test_app/tests.py
```diff
@@ -157,4 +157,3 @@ def test_request_event_admin_no_users(self):
         response = self.client.get(reverse('admin:easyaudit_requestevent_changelist'))
         self.assertEqual(200, response.status_code)
         filters = self._list_filters(response.content)
-        print(filters)
```
jheld, line 160: thank you!

jheld:
released to PyPI for `1.2.3a2`

Issue #62, opened 2018-05-06
Use request.user to log request signal


kunalgrover05:
In the request signals currently, the user is populated using the ```COOKIES``` which wouldn't work for REST APIs based on Django. Can we instead use the ```request.user``` object and make sure that the middleware is included after the Authentication middleware?

https://github.com/soynatan/django-easy-audit/blob/3518a7b3d6225f6b0340c920ebedfad9490eafe1/easyaudit/signals/request_signals.py#L29

chris-brent:
I was having the same problem. My PR is here:
https://github.com/soynatan/django-easy-audit/pull/130

PR #130, opened 2020-03-11
Fix for Issue #62 use request.user in the request signal

chris-brent:


jheld:
thank you. wow, didn't realize we weren't even trying to grab the user (if we could get it via the helper).

jheld:
is it possible to add a test for this "new" behavior?

chris-brent:
> is it possible to add a test for this "new" behavior?

Sure. You do have tests for the requests now? I only see these:
https://github.com/soynatan/django-easy-audit/blob/d772175500004322b096eb134367c5afdc5b06ef/easyaudit/tests/test_app/tests.py#L24

jheld:
@chris-brent I'm unsure what tests are available/around.

But it would be great if you could add something that is able to test this new scenario.

chrisbrent:
Hi I've left IONpath since last updating this so I don't have access to that fork anymore. I've opened a new PR with the same title and one updated test.

PR #134, opened 2020-04-23
Increase admin list panel performance by selecting related

henribru:
The list view in the admin panels is quite slow, especially for the CRUD events. The reason is that it has to do separate queries for the user (and the content type for CRUD events) for each row. This change ensures that all those related objects are fetched in the original query, which gives a pretty massive speed-up.

### DIFF
--- a/easyaudit/admin.py
+++ b/easyaudit/admin.py
@@ -25,14 +25,15 @@ class CRUDEventAdmin(EasyAuditModelAdmin):
                        'object_repr', 'object_json_repr_prettified', 'get_user',
                        'user_pk_as_string', 'datetime', 'changed_fields_prettified']
     exclude = ['object_json_repr', 'changed_fields']
+    list_select_related = ["content_type", "user"]
 
     def get_content_type(self, obj):
-        return ContentType.objects.get(id=obj.content_type_id)
+        return obj.content_type
 
     get_content_type.short_description = "Content Type"
 
     def get_user(self, obj):
-        return get_user_model().objects.filter(id=obj.user_id).first()
+        return obj.user
 
     get_user.short_description = "User"
 
@@ -76,12 +77,12 @@ class LoginEventAdmin(EasyAuditModelAdmin):
     readonly_fields = ['login_type', 'get_username', 'get_user', 'remote_ip', 'datetime', ]
 
     def get_user(self, obj):
-        return get_user_model().objects.filter(id=obj.user_id).first()
+        return obj.user
 
     get_user.short_description = "User"
 
     def get_username(self, obj):
-        user = get_user_model().objects.filter(id=obj.user_id).first()
+        user = obj.user
         username = user.get_username() if user else None
         return username
 
@@ -101,7 +102,7 @@ class RequestEventAdmin(EasyAuditModelAdmin):
     readonly_fields = ['url', 'method', 'query_string', 'get_user', 'remote_ip', 'datetime', ]
 
     def get_user(self, obj):
-        return get_user_model().objects.filter(id=obj.user_id).first()
+        return obj.user
 
     get_user.short_description = "User"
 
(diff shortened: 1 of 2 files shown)


jheld:
I am willing to consider putting this in, but I do want to caution around the following:

- if a project is using split DB, this change will break their use of admin and force them to subclass. (This isn't a major concern, but it is a compatibility breakage).
- - is there a way to emulate the select related improvement while not breaking compatibility? Perhaps getting all user and content type objects as separate querysets based on the crudevent queryset (and either loop through the results for soft linking or make dict of id to object).

henribru:
> * if a project is using split DB, this change will break their use of admin and force them to subclass. (This isn't a major concern, but it is a compatibility breakage).

I'm not familiar with the use of multiple databases in Django. I assume if `CRUDEVENT` is on a different database the `select_related` itself will fail as it tries to join tables that aren't in the same database? Do you know whether `prefetch_related` will work instead, since it does a separate query instead of a join?

>   * is there a way to emulate the select related improvement while not breaking compatibility? Perhaps getting all user and content type objects as separate querysets based on the crudevent queryset (and either loop through the results for soft linking or make dict of id to object).

Yeah, this is doable by overriding `get_changelist_instance` :

```python
    def get_changelist_instance(self, *args, **kwargs):
        changelist_instance = super().get_changelist_instance(*args, **kwargs)
        content_type_ids = [obj.content_type_id for obj in changelist_instance.result_list]
        user_ids = [obj.user_id for obj in changelist_instance.result_list]
        self.content_types_by_id = {content_type.id: content_type for content_type in ContentType.objects.filter(id__in=content_type_ids)}
        self.users_by_id = {user.id: user for user in get_user_model().objects.filter(id__in=user_ids)}
        return changelist_instance
```

> Also would you be willing to share the benchmarks you have? CRUDEvent admin can show up to 100 items so I would be interested to see what speed gains you've had and the number of results returned.

I haven't done a particularly formal benchmark, but the difference I'm seeing is a reduction in the total time as reported by Django Debug Toolbar from around 11 seconds to around 1 second. This is with a full page, i.e. 100 objects. Only one of them has an associated user object, if more of them did the difference would be greater.

jheld:
Yeah I used to refrain from including DDT support in my project. But now we allow it, using the hook/callable on/off flow so we can control more how it's enabled.

All that said, the database time will always be correct (well, resource time). Python and page rendering speed will get slowed down.

Are you saying that the response latency was (most solely) due to DDT, and not really database latency?

henribru:
> Are you saying that the response latency was (most solely) due to DDT, and not really database latency?

Yep. The time reported by DDT for the database is about 270 ms for the current implementation and around 30 ms with this PR. It seems like DDT has a huge overhead for recording queries, which makes it really slow when it has to do that for 309 of them.

If you're still willing to accept this PR despite its utility being a bit more limited than I anticipated I can go ahead and change it to use the `get_changelist_instance` override.

jheld:
Quick question re:11s duration you saw before. Was 11s from debug toolbar, and without debug toolbar it was much faster?

henribru:
Alright, I pushed the change.

easyaudit/admin_helpers.py
```diff
@@ -30,9 +30,14 @@ def prettify_json(json_string):
 
 
 class EasyAuditModelAdmin(admin.ModelAdmin):
+    def get_changelist_instance(self, *args, **kwargs):
+        changelist_instance = super().get_changelist_instance(*args, **kwargs)
+        user_ids = [obj.user_id for obj in changelist_instance.result_list]
```
jheld, line 35: Minor (non-blocking), we could do `user_ids = set(...)`. It won't affect the database, but it may lower the memory usage a _little_. Non-blocking.

easyaudit/admin.py
```diff
@@ -26,13 +26,19 @@ class CRUDEventAdmin(EasyAuditModelAdmin):
                        'user_pk_as_string', 'datetime', 'changed_fields_prettified']
     exclude = ['object_json_repr', 'changed_fields']
 
+    def get_changelist_instance(self, *args, **kwargs):
+        changelist_instance = super().get_changelist_instance(*args, **kwargs)
+        content_type_ids = [obj.content_type_id for obj in changelist_instance.result_list]
```
jheld, line 31: Similar thing here. Maybe good for a downstream change.

jheld:
I've released `1.2.3a1` to PyPI. Currently this is the only change (I'm unsure if I'll add more to this alpha, but either way, it's available!).

Issue #131, opened 2020-03-16
Presave signal fails when using UUID field as PK


thommor:
The pre_save signal fails when a model uses a UUID field as a primary key.

It seems that by the time the pre_save runs the UUID field already has a value meaning the easy audit thinks it is an update and tries to retrieve the original model from the database, which fails.

```
class Customer(ExportModelOperationsMixin('customer'), models.Model):
    """Model that represents a customer"""
    id = models.UUIDField(verbose_name='ID', primary_key=True, default=uuid.uuid4, editable=False, unique=True,
                          help_text='A UUID string identifying this customer')
    """The unique UUID of this :class:`Customer`"""
```

I think this is due to the fact that the UUID field has a default but this is required for UUID fields

nshmn:
I encountered a similar issue when trying to profile my project with django-silk.
Silk logs request with an UUID as id/pk in a varchar field and easy audit tries to store them in an integer field. I got rid of that error with changing the data type of the object_id the field to varchar BUT I didn't check if that causes other problems somewhere else, so be careful!

thommor:
Yes I came to the same conclusion.

Im looking into other packages for the time being until a time non-integer PKs are supported

nshmn:
https://github.com/soynatan/django-easy-audit/pull/103

I just saw that there is already a pull request addressing this.

Issue #90, opened 2019-05-31
Charfield Primary Key


steverecio:
I'm getting a pre-save exception on my models that use a guid charfield as their primary key. This is the line that I believe is causing the issue: https://github.com/soynatan/django-easy-audit/blob/master/easyaudit/signals/model_signals.py#L101

For reference here is a snippet of my model:

```
def _generate_uuid():
    return uuid.uuid4().hex

class SampleModel(models.Model):
    _id = models.CharField(max_length=32, primary_key=True, editable=False, default=_generate_uuid)
```

steverecio:
This can be fixed by changing the object_id on `CRUDEvent` to a CharField: https://github.com/soynatan/django-easy-audit/blob/03e05bc94beb29fc3e4ff86e313a6fef4b766b4b/easyaudit/models.py#L23

See the "Primary key type compatibility" section here: https://docs.djangoproject.com/en/2.2/ref/contrib/contenttypes/#generic-relations

jheld:
We have been hesitant to change to a generic relation. Set null, do nothing semantics etc for deletion.

However yes changing to charfield has been a consideration.

Moving to a charfield should work ok. I think django will auto type convert str IDs to int and vice versa (look ups in either direction).

PR #138, opened 2020-05-14
Change "object_id" to be a "CharField"

jsoa:
The code in this change set migrates the current `CrudEvent.object_id` from an `IntegerField` to a `CharField` as a means to allow various objects with different `PK` types. 

This includes a change to the `model_signals` that query's for the object instead of looking for the `PK`, this is because the common `UUIDField` as a `PK` usually has a `default`, thereofre `PK` would never be empty, even on creation.

This also includes a bug fix for `post_delete` with the `PK` not being available in the `crud_flow` closure. See https://github.com/soynatan/django-easy-audit/pull/138#issuecomment-629869400

Fixes: #131 
Fixes: #90 

- [x] Tested with `Django` `3.0`
- [x] Tested with `UUIDField`
- [x] Tested with `BigAutoField`
- [x] Added Unit tests for `UUIDField` and `BigAutoField` 
- [x] Added Unit tests for deleting objects
- [x] Added Test models for `UUIDField` and `BigAutoField`
- [x] Added test app endpoints for `UUID` and `BigInt`
- [x] Includes updates to the sqlite DB with data related `UUID` and `BigInt`

# Deep dive

The main purpose of these tests is to ensure current users of this app, will not need worry about the migration that will change the type of `object_id` from integer to char. All the output below should be the same per test group, i.e. django version and db. If anything should be different, it will be noted below.

All the test below will do the following

1. Checkout current master and initialize the test project
1. Hit `test_app/create_obj/` 5 times
1. Hit `test_app/update_obj/?id=N` for the 5 items above
1. print the following to verify the current type of object is `int` and basic query works
```shell
>>> for x in CRUDEvent.objects.all():
...     print(x.content_type, x.object_id, type(x.object_id))
...
test_app | test model 5 <class 'int'>
test_app | test model 4 <class 'int'>
test_app | test model 3 <class 'int'>
test_app | test model 2 <class 'int'>
test_app | test model 1 <class 'int'>
test_app | test model 5 <class 'int'>
test_app | test model 4 <class 'int'>
test_app | test model 3 <class 'int'>
test_app | test model 2 <class 'int'>
test_app | test model 1 <class 'int'>
 >>> CRUDEvent.objects.filter(object_id=int(1))
<QuerySet [<CRUDEvent: CRUDEvent object (6)>, <CRUDEvent: CRUDEvent object (1)>]>
```
5. checkout the new branch and run migrations
6. print the same output as above, ensuring the new type is `str` and the query still works and returns the same results, by passing it str and int versions.
```shell
>>> for x in CRUDEvent.objects.all():
...     print(x.content_type, x.object_id, type(x.object_id))
...
test_app | test model 5 <class 'str'>
test_app | test model 4 <class 'str'>
test_app | test model 3 <class 'str'>
test_app | test model 2 <class 'str'>
test_app | test model 1 <class 'str'>
test_app | test model 5 <class 'str'>
test_app | test model 4 <class 'str'>
test_app | test model 3 <class 'str'>
test_app | test model 2 <class 'str'>
test_app | test model 1 <class 'str'>
>>> CRUDEvent.objects.filter(object_id='1')
<QuerySet [<CRUDEvent: CRUDEvent object (6)>, <CRUDEvent: CRUDEvent object (1)>]>
>>> CRUDEvent.objects.filter(object_id=int(1))
<QuerySet [<CRUDEvent: CRUDEvent object (6)>, <CRUDEvent: CRUDEvent object (1)>]>

```
## Current progress

|    | Django Version | DB       | DB Version |
| ---- | ---------------- | ---------- | ------------ |
| [x]  |            3.0 | Postgres |       12.0 |
| [x]  |            3.0 | Postgres |       11.0 |
| [x]  |            3.0 | Postgres |       10.0 |
| [x]  |            3.0 | Postgres |          9 |
| [x]  |            2.2 | Postgres |       12.0 |
| [x]  |            2.2 | Postgres |       11.0 |
| [x]  |            2.2 | Postgres |       10.0 |
| [x]  |            2.2 | Postgres |        9.0 |
| [x]* |           1.11 | Postgres |       12.0 |
| [x]* |           1.11 | Postgres |       11.0 |
| [x]* |           1.11 | Postgres |       10.0 |
| [x]* |           1.11 | Postgres |        9.0 |
| [x] |            3.0 | MySQL    |          8 |
| [x] |            3.0 | MySQL    |          5 |
| [x] |            2.2 | MySQL    |          8 |
| [x] |            2.2 | MySQL    |          5 |
| [x]* |           1.11 | MySQL    |          8 |
| [x]* |           1.11 | MySQL    |          5 |

### Notes
- Django 1.11 created more `CRUDEvent` objects (on master branch)
```shell
>>> for x in CRUDEvent.objects.all():
...     print(x.content_type, x.object_id, type(x.object_id))
...
test model 5 <class 'int'>
test model 4 <class 'int'>
test model 3 <class 'int'>
test model 2 <class 'int'>
test model 1 <class 'int'>
test model 5 <class 'int'>
test model 4 <class 'int'>
test model 3 <class 'int'>
test model 2 <class 'int'>
test model 1 <class 'int'>
content type 12 <class 'int'>
content type 11 <class 'int'>
content type 10 <class 'int'>
content type 9 <class 'int'>
content type 8 <class 'int'>
content type 7 <class 'int'>
content type 6 <class 'int'>
content type 5 <class 'int'>
content type 4 <class 'int'>
content type 3 <class 'int'>
content type 1 <class 'int'>
```
- After migrating
```shell
content type 18 <class 'str'>
content type 17 <class 'str'>
content type 16 <class 'str'>
content type 15 <class 'str'>
content type 14 <class 'str'>
content type 13 <class 'str'>
test model 5 <class 'str'>
test model 4 <class 'str'>
test model 3 <class 'str'>
test model 2 <class 'str'>
test model 1 <class 'str'>
test model 5 <class 'str'>
test model 4 <class 'str'>
test model 3 <class 'str'>
test model 2 <class 'str'>
test model 1 <class 'str'>
content type 12 <class 'str'>
content type 11 <class 'str'>
content type 10 <class 'str'>
content type 9 <class 'str'>
content type 8 <class 'str'>
content type 7 <class 'str'>
content type 6 <class 'str'>
content type 5 <class 'str'>
content type 4 <class 'str'>
content type 3 <class 'str'>
content type 1 <class 'str'>
```


easyaudit/signals/model_signals.py
```diff
@@ -62,14 +62,16 @@ def pre_save(sender, instance, raw, using, update_fields, **kwargs):
                 # We need a better way for this to work. ManyToMany will fail on pre_save on create
                 return None
 
-            if instance.pk is None:
-                created = True
-            else:
+            # Query the object, see if it exists
+            try:
+                old_model = sender.objects.get(id=instance.pk)
```
steverecio, line 67: Please change this to `pk=instance.pk`

I have primary keys not named `id` so this throws an exception

jsoa: @steverecio updated

steverecio:
One note after testing this a bit more. The `object_id` should have null=True to handle Delete events. I'm seeing the following error after deleting objects:

```
easy audit had a post_delete exception on CRUDEvent creation
...
IntegrityError: null value in column "object_id" violates not-null constraint
```

jsoa:
@steverecio I was seeing this as well with postgres, doesn't seem to be an issue with sqlite. 

The instance id is stripped before the transaction completes with latest postgres. Simple fix i think.

jsoa:
Last commit should address the issue without needing to make the field nullable, here is a screenshot of a project ive been testing with..

steverecio:
@jsoa nice fix! I like your approach much better.

@jheld @soynatan can we get this merged in?

easyaudit/signals/model_signals.py
```diff
@@ -62,14 +62,16 @@ def pre_save(sender, instance, raw, using, update_fields, **kwargs):
                 # We need a better way for this to work. ManyToMany will fail on pre_save on create
                 return None
 
-            if instance.pk is None:
```
Etenil, line 65: I wonder if there are performance considerations here.

I've made a similar change in a PR but I opted to keep the `if instance.pk`. My logic was that if `pk` is None it's reasonable to assume that the object does not exist in database, this way a database request can be saved.

I don't know what way is best, hopefully more experienced people than me can chime in.

jsoa: I tend to agree and i think your approach is better than this. either way the performance is still impacted with a query.

jsoa: @Etenil Added your changes from #141

jsoa:
@Etenil @jheld Updated code in order to not do the explicit query in the pre_save signal. Instead it will use the internal `_state` object to determine if the object is being added. Took a look and `_state` has been around since django 1.2

jheld:
Oh that's a nice idea.

We use the `_state` a little in my project (I don't love it -- I wish django would make the contract exposed/official). I think so long as our test case is written such that if that goes away/changes, then it'll fail and then we'll be aware before release.

I/we need to first fix/address the breakage on the request event user signal before merging anything else. Then we can add a couple more things (nothing which breaks compatibility/super risky).

This seems like a good candidate!

jsoa:
@jheld I also don't love checking `_state`, however I did add a compatibility test to ensure its available.

jheld:
Excellent work with all of the test cases and test models. Comprehensive and pretty modular/DRY.

This looks like a great candidate for `1.3.0`. I hope to finalize `1.2.3` this or next week.

jheld:
Likely will release `1.2.3` this week/early next week. There was a bug introduced in the request event handler at the beginning of the alphas and the last/current alpha (4) effectively rolls back that change.

jheld:
About ready to merge this!

@jsoa 

So `TestModel` is still implicitly using `AutoField` (e.g. regular INT), right? And that is the "base" test case model so it is tested, and that's good.

Can you take a look over the comments/reviews once more to make sure all of those are answered/moot?

jsoa:
@jheld That is correct `TestModel` is the basic django model with `AutoField` (integer PK)

I took a look again and everything seems good to go. The only thing that i'm not 100% certain of is if existing users of this app will have to do anything special after the schema migration is run, such as converting the integers to chars.

jheld:
@jsoa if I'm understanding your question, so long as they run migrations, they shouldn't have to do anything other than that. I'm happy to test on my local, too, next week. My project runs on the default configuration.

pcraig3:
Hello! Really excited to see this PR go in. 

I'm working on a Django project where we're using UUIDs as primary keys for our users and this PR resolves the issues I'm seeing. It seems like it's nearly there, so hope this can be merged soon! 👍

jsoa:
@jheld started testing the migration (manually). See PR description for details, still work in progress. we can add more DB versions and/or django versions if needed.

jheld:
@jsoa this looks really great. Thank you for running through some simple (but time consuming!) test cases. My project uses MySQL but if you think you'll have time to run through that on your end I'll wait til then. Let me know -- happy to follow the instructions you've laid out.

jsoa:
@jheld done with my testing, everything seems good to go, with the minor note about django 1.11. It would be nice to get someone to run through just some of the manual tests that was run.

Here is a docker-compose file i used for bootstrapping the DBs
```
version: "3"

services:

  # POSTGRES
  ea-psql12:
    image: postgres:12.0

    environment:
      - POSTGRES_PASSWORD=ea
      - POSTGRES_USER=ea
      - POSTGRES_DB=ea

    ports:
      - "54320:5432"

  ea-psql11:
    image: postgres:11.0

    environment:
      - POSTGRES_PASSWORD=ea
      - POSTGRES_USER=ea
      - POSTGRES_DB=ea

    ports:
      - "54321:5432"


  ea-psql10:
    image: postgres:10.0

    environment:
      - POSTGRES_PASSWORD=ea
      - POSTGRES_USER=ea
      - POSTGRES_DB=ea

    ports:
      - "54322:5432"

  ea-psql9:
    image: postgres:9

    environment:
      - POSTGRES_PASSWORD=ea
      - POSTGRES_USER=ea
      - POSTGRES_DB=ea

    ports:
      - "54323:5432"


  # MYSQL
  ea-mysql8:
    image: mysql:8

    environment:
      - MYSQL_ROOT_PASSWORD=ea
      - MYSQL_DATABASE=ea
      - MYSQL_USER=ea
      - MYSQL_PASSWORD=ea

    ports:
      - "33060:3306"

  ea-mysql5:
    image: mysql:5

    environment:
      - MYSQL_ROOT_PASSWORD=ea
      - MYSQL_DATABASE=ea
      - MYSQL_USER=ea
      - MYSQL_PASSWORD=ea

    ports:
      - "33061:3306"
```
Each DB has a different host port
- PG
  - 54320 - 12
  - 54321 - 11
  - 54322 - 10
  - 54323 - 9
- MYSQL
  - 33060 - 8
  - 33061 - 5

After each django version check, we need to reset the containers, by `docker rm <container id>`
Then `docker volume rm $(docker volume ls --quiet --filter dangling=true) || true ` **(removes any volume not attached to a container)**
Then its just manually changed the DB connection in the `test_project` per DB version and Django version.

```python
    # 'default': {
    #     'ENGINE': 'django.db.backends.postgresql',
    #     'NAME': 'ea',
    #     'USER': 'ea',
    #     'PASSWORD': 'ea',
    #     'HOST': 'localhost',
    #     'PORT': '54323'
    # }

    'default': {
        'ENGINE': 'django.db.backends.mysql',
        'NAME': 'ea',
        'USER': 'ea',
        'PASSWORD': 'ea',
        'HOST': '127.0.0.1',
        'PORT': '33061'
    }
```

jheld:
@jsoa excellent use for docker-compose.

We don't support django 1.11 officially anymore.

2.0+ (and realistically other than travis tests passing, we really only need to be 2.2+).

So while I agree 1.11 is odd, I will argue we should ignore the difference. I have no idea what has changed/subtly broken in our usage.

PR #140, opened 2020-05-18
Allow dynamic model registration

Etenil:
The built-in `User` model can't be fetched by a `apps.get_model()` nor given as class as they would go in a recursive import.

```py
# settings.py
from django.contrib.auth import get_user_model() # <= recursive import of settings!
```

This allows to do the following instead:

```py
# settings.py
def lazy_user_model():
    from django.contrib.auth import get_user_model()
    return get_user_model()

DJANGO_EASY_AUDIT_REGISTERED_CLASSES = [
    lazy_user_model
]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Below are some of the settings you may want to use. These should be defined in y
 
   A list of Django models which will be ignored by Django Easy Audit.
   Use it to prevent logging one or more of your project's models.
-  List items can be classes or strings with `app_name.model_name` format.
+  List items can be classes, strings with `app_name.model_name` format or functions that will return classes.
 
 * `DJANGO_EASY_AUDIT_UNREGISTERED_URLS_EXTRA`
 
--- a/easyaudit/settings.py
+++ b/easyaudit/settings.py
@@ -1,5 +1,5 @@
-
 from importlib import import_module
+from inspect import isfunction
 
 import django.db.utils
 from django.apps import apps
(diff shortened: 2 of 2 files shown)


jheld:
thank you for this. Can you please add at least one test that utilizes this?

Etenil:
@jheld I had trouble testing this, as the lazy invocation for the dynamic model happens when the settings are loaded. So `override_settings` won't actually re-run the settings loading logic, it only overrides the pre-loaded settings.

Instead I've opted to pre-register the models in `settings.py` using both the dynamic and string registrations. Commenting any of them out causes the tests to fail.

Let me know if this is good enough or if you have a better idea on how to proceed.

jheld:
@Etenil I appreciate the effort you've put in.

We might merge as is.

I had a feeling this might happen. I think I got around it with the delta/changed fields setting by instead of importing the setting variable directly, doing a getattr on settings when doing the logic. Would that work for you? I think override_settings still works in that case -- and I think/hope I created a test case/altered a test case for it. If that doesn't work/taking too much time to implement, we'll merge as is.

Etenil:
@jheld I'm not really sure how to do what you mentioned, would you have an example?

jheld:
Yes. I think I have both a test case (and supporting code in the library which is being tested) for the delta/changed field on/off flag. The setting that's being tested is in the README. It was I think one of the last betas of the current official release.

If you have trouble finding it let me know I can add more detail to here.

jheld:
@Etenil can you rebase? We may not _require_ an additional test, but want to make sure things are still passing.

PR #141, opened 2020-05-18
Better handle transactions in pre-save

Etenil:
The pre-save hook makes the assumption that a PK that is set denotes a model update. In transactions it can happen that a PK is defined however the model isn't actually saved yet. The current code causes an error in these conditions.


jheld:
Interesting fix. Looks good.

I never really liked the way the delta/changed fields thing worked. I realize it wasn't the easiest code change to make at the time (and swapping to an alternative implementation isn't trivial). But this looks like a good fix, thank you.

Can you also add a test case?

Etenil:
@jheld added test

jheld:
@Etenil can you check if the `pre_save` logic is still in need of change based on current master?

PR #144, opened 2020-05-18
Fix 105

steverecio:
Fixes previous test regression


jheld:
thank you.

It's now up on PyPI as `1.2.3a3`

Issue #81, opened 2019-03-21
request_started_handler() missing 1 required positional argument: 'environ'


dopatraman:
When I include easy_audit via the steps in the readme (add it to INSTALLED_APPS and MIDDLEWARE), I get this error on app startup:
```
request_started_handler() missing 1 required positional argument: 'environ'
```

As far as I can tell, this argument is supposed to be passed to the signal receiver automatically. I'm not sure what's causing this issue.

jheld:
Looks like this was before easyaudit upgraded to have django 1.8+ support. If you upgrade to a more recent version I think you'll find that it works.

CharleneCheung:
import logging
from django.core.signals import request_started
from easyaudit import settings
from easyaudit.signals.request_signals import request_started_handler
logger = logging.getLogger(__name__)
"""
This file override the signal from channels 2.x.
We found a error that when we set DJANGO_EASY_AUDIT_WATCH_REQUEST_EVENTS as True, 
python pop error as below:
request_started_handler() missing 1 required positional argument: 'environ'

In django signal library, it only accept parameter "environ". 
Moreover, channels pass parameter "scope" instead of "environ". 

if you need to use channels as websocket, 
you must rewrite a new method and assign attribute that easy audit needs. 

notes:
In channels:
signals.request_started.send(sender=self.__class__, scope=self.scope)

In django signal
request_started_handler(sender, environ, **kwargs):
"""

def ready():
    print("game signals ready")


def fixed_request_started_handler(sender, scope, **kwargs):
    # TODO replicate environ
    headers = dict(scope['headers'])
    scope['PATH_INFO'] = scope['path']
    scope['HTTP_COOKIE'] = headers.get(b'cookie').decode('utf-8')
    scope['REQUEST_METHOD'] = scope['method']
    scope['QUERY_STRING'] = scope['query_string']
    scope['REMOTE_ADDR'] = scope['client'][0]
    return request_started_handler(sender, environ=scope, **kwargs)


if settings.WATCH_REQUEST_EVENTS:
    dispatch_uid = 'easy_audit_signals_request_started'
    request_started.disconnect(request_started_handler, dispatch_uid=dispatch_uid)
    request_started.connect(fixed_request_started_handler, dispatch_uid=dispatch_uid)

jheld:
@CharleneCheung 

[...]

PR #147, opened 2020-06-06
ASGI RequestEvent support

jheld:
I have also created a test which utilizes ASGI (the view being called is synchronous, but that's fine!). Testing websocket at some point would be nice, as well as async views (websockets can be synchronous, which is a pattern django channels offers as a concrete solution). But yes, websockets likely would require django channels (not opposed to making it an extra install which is grabbed by the test runner), and that's outside the scope of this code change as that's not truly supported in django at this time (not natively).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,12 +10,14 @@ jobs:
       max-parallel: 4
       matrix:
         python-version: [3.5, 3.6, 3.7, 3.8]
-        django-version: [2.2, 2.2.8, 3.0]
+        django-version: [2.2, 2.2.8, 3.0, 3.1]
         exclude:
           - python-version: 3.5
             django-version: 2.2.8
           - python-version: 3.5
             django-version: 3.0
+          - python-version: 3.5
+            django-version: 3.1
           - python-version: 3.8
             django-version: 2.2
 
@@ -28,7 +30,6 @@ jobs:
     - name: Install Dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install django==${{ matrix.django-version }}
         python -m pip install -e .
     - name: Run Tests
       run: |
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,9 @@ env:
     - DJANGO="django>=2.2,<2.2.8"
     - DJANGO="django>=2.2.8,<3.0"
     - DJANGO="django>=3.0,<3.1"
+    - DJANGO="django>=3.1,<3.2"
 
 install:
-    - pip install ${DJANGO}
     - pip install -e .
 
 before_script:
@@ -28,6 +28,8 @@ matrix:
   exclude:
     - python: "3.5"
       env: DJANGO="django>=3.0,<3.1"
+    - python: "3.5"
+      env: DJANGO="django>=3.1,<3.2"
     - python: "3.8"
       env: DJANGO="django>=2.0,<2.1"
     - python: "3.8"
--- a/easyaudit/middleware/easyaudit.py
+++ b/easyaudit/middleware/easyaudit.py
@@ -1,8 +1,3 @@
-try:
-    from django.utils.deprecation import MiddlewareMixin
-except ImportError:
-    # not required in <= 1.9
-    MiddlewareMixin = object
 
 # makes easy-audit thread-safe
 try:
@@ -16,15 +11,19 @@ class MockRequest(object):
         self.user = user
         super(MockRequest, self).__init__(*args, **kwargs)
 
+
 _thread_locals = local()
+
 
 def get_current_request():
     return getattr(_thread_locals, 'request', None)
+
 
 def get_current_user():
     request = get_current_request()
     if request:
         return getattr(request, 'user', None)
+
 
 def set_current_user(user):
     try:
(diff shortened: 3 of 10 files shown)


jheld:
This is designed to fix issue #81

jheld:
@dopatraman would you willing to test this if you're still using the repo w/ASGI?

jheld:
@CharleneCheung same question as above.

I am not entirely sure how to write tests for this yet but I do think it's likely going to work.

CharleneCheung:
or maybe i just do a test later. thank you so much

Jason Held <notifications@github.com> 於 2020年7月25日 週六 上午5:21寫道：

saengate:
@jheld @CharleneCheung Hello, I have an application that uses this and `manage runserver` throws the error in http while websocket works, in nginx http and websocket work perfectly. Disabling easyaudit is what worked for me to work with `manage.py runserver` and debug. I testing this branch with `pip install -e git+https://github.com/soynatan/django-easy-audit.git@feature/asgi-request-event#egg=django-easy-audit`. All work perfectly! I hope that you can mix this soon.

jheld:
Oh that's great!

I am open to merging it without explicit unit tests, but I am a bit reluctant to do so unless we find that short term we don't know how to do it, haha.

But it was working for you, and that appears to give us the realistic confidence!

If you don't hear back within a week, please ping me here.

jheld:
@saengate @CharleneCheung I've taken a swing at writing up a test. Turns out we didn't really have any tests (other than the skipped one which unfortunately had a bug around it so it had to be skipped since the code itself was commented out) for request event itself. We know have a WSGI and ASGI test to help keep this stable. I ran into a lot of confusion on the ASGI/asyncio story on how to do some of this. This is the first time I've really written any async code and even this wasn't that much. However, the tests do pass.

Also, I did have to update the easy audit middleware (well _had_ to is not accurate), so now it's no longer using the `MiddlewareMixin`. I did this partly because we don't want to (at this time) support ASGI so not using the mixin means that django will say "ok, you didn't say, therefore WSGI only", while the mixin had wsgi & asgi support, whoops! It does make the mixin a little more complex, but it may have been more complex with continuing to use the (soon to be deprecated) mixin because of how it logics on asgi and wsgi, and we need to manage the thread locals anyway.

For WSGI only, the code was fine as it was (adding a test was just a bonus). But for ASGI (regardless if the view is async or not), the middleware had to have a change no matter what. So, this is slightly more change than required, but I didn't do more than I felt was needed/warranted.

Would either of you be willing to test this? If so and it looks good, then I'm happy to merge it. If not, then I'll still merge -- the tests pass and we should be comfortable with the tests. That's also why this release is currently still an alpha as things can happen, this is when we find [the big] issues.

saengate:
@jheld something in
[5438de3](https://github.com/soynatan/django-easy-audit/pull/147/commits/5438de357bcd69cc9bdefed8a5c090f8c477f563) has failed the application again, the same test I had performed before is now failing. If it helps you, i share my project, is open source [djfullapp](https://github.com/saengate/djfullapp) you can test this using `./manage.py runserver 0.0.0.0:7000` in the container and the port `7003` in localhost. `localhost:7003`

jheld:
@saengate oh no! So, the only things that really changed were that i added a string `decode` line in the request started handler, and that the middleware now is more explicitly written, rather than relying on the mixin. Hmm.

saengate:
@jheld try the previous commit and it also fails now, there is some forced change in this which is the cause of the error again probe `pip install -e git+https://github.com/soynatan/django-easy-audit.git@e7b80c3eb9f6ed75f28c922f5eee585d44dd6b24#egg=django-easy-audit`

jheld:
@saengate so commit [e7b80](https://github.com/soynatan/django-easy-audit/pull/147/commits/e7b80c3eb9f6ed75f28c922f5eee585d44dd6b24) , the initial commit, is now not working? I didn't change that commit. I've only been building on top of it.

jheld:
Also, do you have any errors/tracebacks which you can share?

Also, anytime we have to re-test this branch, we probably will want to first pip uninstall before installing again.

saengate:
@jheld oh! My mistake was that I was not using the correct settings for the test. I testing the branch again and it's ok.
this time I ran with a
`pip install -e git+https://github.com/soynatan/django-easy-audit.git@feature/asgi-request-event#egg=django-easy-audit`
`./manage.py runserver 0.0.0.0:7000 --settings=config.debug.settings`
`pip freeze | grep django-easy-audit
-e git+https://github.com/soynatan/django-easy-audit.git@6c435ee01b1dd282e3bf9d20da55a73a5c15ec2a#egg=django_easy_audit`

jheld:
@saengate no problem! Thank you for testing & providing feedback. Ok, I'll merge this and get it out in the `a4`.

jheld:
OK, `1.3.0a4` is now on PyPI.

We'll give it a little bit of testing time, and then we'll officially release 1.3.0.

jheld:
@saengate are you feeling confident with this version? I think this is the last change for the release. I'm happy to officially release `1.3.0` this week/end.

saengate:
@jheld I had been a bit detached from this, but this week I will be using the project more thoroughly, if it has not been yet mixed and find any problem related this, i comment you.

jheld:
I likely will cut the official release this weekend or early next week. Between the CharField object_id improvement, fixing a bug in a transaction on pre save (I think), and this, we're full with enhancements and we'll make any hotfixes an necessary. We've had this alpha out for several weeks so folks have had time to raise any concerns.

jheld:
`1.3.0` is now on PyPI.

PR #149, opened 2020-06-19
check if postgresql to fix 105

steverecio:


jheld:
Thank you! `1.2.3a5` is now on PyPI. The intention is that this is the last true release for `1.2.3`. I may cut a rc tag after `a5` but we'll see. I'll be sending this up through my company's SDLC and once it passes through QA sometime next week ideally, then we're set to cut the official release, and then move onto more fun merges for the next release!

PR #162, opened 2020-09-28
Make models translatable and add Russian translation

Real-Gecko:


easyaudit/models.py
```diff
@@ -1,6 +1,7 @@
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from django.db import models
+from django.utils.translation import ugettext_lazy as _
```
jheld, line 4: I think django is deprecating this in favor of `gettext_lazy`. Can you look into it before merge?

jheld (changes requested):
Just a small thing about a django import, looks good otherwise, thank you!

easyaudit/models.py
Real-Gecko: Yes you're right `ugettext_lazy` is deprecated and is alias for `gettext_lazy`, commited changes.

jheld (approved):

jheld:
Probably won't cut a release just yet (depends on your other PR and any others), but likely will within the next 2 or 3 weeks. Please ping me if I haven't.

Real-Gecko:
There's one more PR worth checking #163

PR #168, opened 2020-11-25
_base_manager instead of objects

gurkankisaoglu:
In our project default object managers are overridden in most of the models and we use the django-easy-audit repository in several places, this sometimes creates errors. 
Error scenario:
- We have User model and default object manager overridden for excluding inactive users from queryset.
- Making change any model instance that a deactivated user is connected to triggers easy audit’s pre save signal to audit the change. 
-Easy audit is trying to log the change by creating a CRUDEvent instance but since we exclude inactive users in default object manager and easy audit uses object in their pre save signal, user is not found.

### DIFF
--- a/easyaudit/signals/model_signals.py
+++ b/easyaudit/signals/model_signals.py
@@ -68,7 +68,7 @@ def pre_save(sender, instance, raw, using, update_fields, **kwargs):
 
             # created or updated?
             if not created:
-                old_model = sender.objects.get(pk=instance.pk)
+                old_model = sender._base_manager.get(pk=instance.pk)
                 delta = model_delta(old_model, instance)
                 if not delta and getattr(settings, "DJANGO_EASY_AUDIT_CRUD_EVENT_NO_CHANGED_FIELDS_SKIP", False):
                     return False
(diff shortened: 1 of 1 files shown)


jheld:
Before I review this, could you check if the easy audit callback hook is used in pre-save? If so, I'd recommend you use that. If not...would adding it to the pre-save work?

bugraaydin:
> Before I review this, could you check if the easy audit callback hook is used in pre-save? If so, I'd recommend you use that. If not...would adding it to the pre-save work?

Hello @jheld could you elaborate `easy audit callback hook` please?

jheld:
Sure. There should be some references to: `DJANGO_EASY_AUDIT_CRUD_DIFFERENCE_CALLBACKS`

gokberk-sahin:
Hello @jheld! The problem that we are having is an exception caused by we overriding the object manager to filter out the deactivated users. This causes the `old_model = sender.objects.get(pk=instance.pk)` throw a `DoesNotExist` exception before even checking for the callbacks. So I propose that we make the callback checks and return `False` early if necessary in #172 . I would appreciate if you take a look

JirkaV:
Incidentally I didn't notice this and created https://github.com/soynatan/django-easy-audit/pull/176 - the only difference is that I didn't feel the need to use `_base_manager` for Django ContentType.

PR #172, opened 2020-12-31
Move the callback checks in model_signals to the beginning

gokberk-sahin:
Hey @jheld, to resolve the #168 I propose this solution. However I was wondering why the callback are not getting checked in other signals?


gokberk-sahin:
Hey @jheld! Can you take a look at this when you have the time please?

PR #176, opened 2021-02-06
Fix for Issue 175 (possible exception with custom managers hiding som…

JirkaV:
As promised - use Django `_default_manager` to prevent exceptions when custom manager excludes certain model instances.

### DIFF
--- a/easyaudit/signals/model_signals.py
+++ b/easyaudit/signals/model_signals.py
@@ -68,7 +68,7 @@ def pre_save(sender, instance, raw, using, update_fields, **kwargs):
 
             # created or updated?
             if not created:
-                old_model = sender.objects.get(pk=instance.pk)
+                old_model = sender._default_manager.get(pk=instance.pk)
                 delta = model_delta(old_model, instance)
                 if not delta and getattr(settings, "DJANGO_EASY_AUDIT_CRUD_EVENT_NO_CHANGED_FIELDS_SKIP", False):
                     return False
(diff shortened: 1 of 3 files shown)


easyaudit/signals/model_signals.py
```diff
@@ -68,7 +68,7 @@ def pre_save(sender, instance, raw, using, update_fields, **kwargs):
 
             # created or updated?
             if not created:
-                old_model = sender.objects.get(pk=instance.pk)
+                old_model = sender._default_manager.get(pk=instance.pk)
```
klplentific, line 71: _base_manager i think

jheld: https://docs.djangoproject.com/en/4.0/topics/db/managers/#default-managers-1, https://docs.djangoproject.com/en/4.0/topics/db/managers/#base-managers

can someone verify the most correct solution?

jheld (changes requested):

sgordon16:
`_base_manager` will return all records, the `_default_manager` can still be filtering out records and might still raise an DoesNotExist exception.
[https://docs.djangoproject.com/en/5.0/topics/db/managers/#django.db.models.Model._base_manager](https://docs.djangoproject.com/en/5.0/topics/db/managers/#django.db.models.Model._base_manager)

rgbanasiak:
Is there a chance that the pull request will be accepted?

easyaudit/signals/model_signals.py
sgordon16: It should be _base_manager which will always return all records, the _default_manager may be filtering out records and can still raise an DoesNotExist exception.

tarasinf:
Any progress with this?

Issue #165, opened 2020-11-03
No user_id is recorded


hasahmad:
I am using this with Django rest framework, using OAuth2, but it's not recording `user_id`
Are you only checking for users in session cookie?

jheld:
I think that's a fairly accurate statement. Will accept PR to change that. We do support general usage of the login_user signal (e.g. manually telling Django, and I've seen that work for JWT login via DRF hooks). If there is a way for the library to more seamlessly hook into the general system, that would be good.

wmantly:
@hasahmad did you find a workaround? I am having the same issue?

zoot-io:
@hasahmad @wmantly 
I have the same issue, did you figure it out?

wmantly:
@hasahmad, I posted this on behalf of one of my clients. The issue was due to him not authenticating some requests. Simply fixing some calls he was making anonymously fixed this issue(and a security hole he had)

zoot-io:
> @hasahmad, I posted this on behalf of one of my clients. The issue was due to him not authenticating some requests. Simply fixing some calls he was making anonymously fixed this issue(and a security hole he had)

What do you mean by that '... not authenticating some requests. Simply fixing some calls ...'? My project uses a lot of interconnected models and most (if not all) are exposed through authenticated ViewSets using the DRF's `IsAuthenticated` permission class. 

Are you able to provide some detail about what he did specifically?

mschoettle:
Which DRF authentication classes are you using in your project?

mschoettle:
I just tested it with `TokenAuthentication` and it does not record the user. I think it is due to the fact that DRF does not plug their authentication mechanisms into Django's authentication system (it works with `SessionAuthentication` because it uses the standard Django session). I noticed recently that when doing a request with `TokenAuthentication`, the user is `AnonymousUser` (see [my comment here](https://github.com/encode/django-rest-framework/issues/8359#issuecomment-1064390980)). Not totally sure whether this is a bug in DRF or not.

mschoettle:
Turns out that there are already some solutions, unfortunately in DRF so everyone could benefit from them. One of them is DRF JWT: https://github.com/jpadilla/django-rest-framework-jwt/issues/45#issuecomment-467576074

Basically, DRF authenticates in the view layer. The workaround is to supply a custom middleware that does the authentication. However, it leads to authenticating twice and thus increasing the number of DB queries.

[...]

PR #178, opened 2021-03-02
Get session using SessionStore

tomasgarzon:
Allow getting the user from the cache, not only from django-db. Using SessionStore abstraction.

### DIFF
--- a/easyaudit/signals/request_signals.py
+++ b/easyaudit/signals/request_signals.py
@@ -5,6 +5,7 @@ from django.http.cookie import SimpleCookie
 from django.utils import timezone
 from django.conf import settings
 from django.utils.module_loading import import_string
+from importlib import import_module
 
 # try and get the user from the request; commented for now, may have a bug in this flow.
 # from easyaudit.middleware.easyaudit import get_current_user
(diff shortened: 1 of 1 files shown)


easyaudit/signals/request_signals.py
```diff
@@ -69,14 +70,12 @@ def request_started_handler(sender, **kwargs):
         session_cookie_name = settings.SESSION_COOKIE_NAME
         if session_cookie_name in cookie:
             session_id = cookie[session_cookie_name].value
-
-            try:
-                session = Session.objects.get(session_key=session_id)
-            except Session.DoesNotExist:
-                session = None
+            engine = import_module(settings.SESSION_ENGINE)
```
jheld, line 73: I don't particularly mind this flow, but is there a less brutal way of doing this for users that do not require as much flexibility in the call flow? Maybe a new settings flag?
Also, is there harm in resolving the `engine` in the module scope, as opposed to the function scope? (Perhaps there is, like a django init thing, but wanted to ask just in case there was no danger.)

jheld (commented):
Just some thoughts and look forward to your insight.

CleitonDeLima:
I decided to test django-easy-audit and I liked this project.
I had this problem, my sessions are saved in redis and not in django-db. Do you have any predictions for this fix coming in?

easyaudit/signals/request_signals.py
mschoettle: Are you referring to the use of `import_module`? It's the recommended way in the [Django docs](https://docs.djangoproject.com/en/dev/topics/http/sessions/#using-sessions-out-of-views). If the user is retrieved from `request.user` this would not be required at all actually (see my comment in #165).

I agree that the session store should be retrieved once instead of each request. Maybe move it to the top of the module where `audit_logger` is defined?

Issue #177, opened 2021-03-02
Obtaining user_pk_as_string from request during creation of CRUDEvent if an header is present


bugraaydin:
Hello! First of all thank you for maintaining this amazing project, it is really helpful. I wanted to ask something. In my project, User model is in another microservice. f/e sends `x-user-uuid` header while doing CRUD operations. I want to be able to set `user_pk_as_string` to this value, if it is present during CRUDEvent creation (both User and user_pk_as_string are None in my case as user model is in another service and a session is used in my service). Is it possible, if yes, how can I do it?

Also, is there an example usage of `DJANGO_EASY_AUDIT_CRUD_DIFFERENCE_CALLBACKS` which I can take a look to further add some configuration to logging events. 

Thank you

PR #179, opened 2021-03-02
Functionality to set user_pk_as_string without a dependency on User fk

bugraaydin:
Implementation to log user information is completely dependent on User fk. This makes it really hard to use this repository for applications that has User model in a different service than the current service.

addresses https://github.com/soynatan/django-easy-audit/issues/177


bugraaydin:
Hey @jheld Could you please review this PR if possible. It adds ability to store user_pk_as_string field without using a User instance, while keeping backwards compatibility

Issue #182, opened 2021-04-08
Django 3.2 support


tian-yi:
Thanks for this great package.

Django made 3.2 release a couple of days ago.

I wonder if there's any plan to upgrade django-easy-audit to be compatible with it?

Thanks!

serge-gaia:
Django 3.2 being the new LTS, it's important.

jheld:
Do either of you have time to create a PR for this?

tian-yi:
I will give it a try next week.

pembo13:
Are there known incompatibilities with Django 3.2?

Real-Gecko:
Tried to upgrade too, tests fail with
`django.db.utils.OperationalError: database table is locked: django_session` with Django 3.2 starting from Python 3.6 up to 3.9.

tian-yi:
Yeah, saw this error too. I guess my simple attempt to just upgrading the `setup.py` does not work. Need more work for this to work with Django 3.2 I think.

Real-Gecko:
Actually all tests pass except one `async def test_login`, actual error happens on this line:
https://github.com/soynatan/django-easy-audit/blob/860a556d0f9314e25d99904643c2a4e18c71aa51/easyaudit/tests/test_app/tests.py#L211
I guess some ASGI features changes in 3.2.
By the way they also introduced the [DEFAULT_AUTO_FIELD](https://docs.djangoproject.com/en/3.2/releases/3.2/#customizing-type-of-auto-created-primary-keys) setting, so `easyaudit/tests/test_project/settings.py` is better to be updated to avoid annoying hints https://github.com/Real-Gecko/django-easy-audit/commit/8811dff8376ba869847a52e6c1519c42c316de70#

jkarstens:
The [async `test_login`](https://github.com/soynatan/django-easy-audit/blob/860a556d0f9314e25d99904643c2a4e18c71aa51/easyaudit/tests/test_app/tests.py#L208) test failure isn't specific to Django 3.2 (it fails on Django 3.1.8 as well). 

> [Note that sync code will always be in a different thread to any async code that is calling it](https://docs.djangoproject.com/en/3.2/topics/async/). 

Since each `TestCase` test is wrapped in a transaction, this means multiple threads are accessing the same database tables within a transaction (hence the `database table is locked` error). 

[...]

PR #184, opened 2021-04-25
Add support for Django 3.2

tian-yi:
Add Django 3.2 to supported version.
Fix issue #182 

Hi @jheld,
I cannot find a contribution guide in this project, so I am not sure if I've done the right thing. :)


tian-yi:
Close this, tests failing for Django 3.2. By simply updating the `setup.py` file would not work. 
I will try to dig a bit deeper to see what the problem is.

PR #185, opened 2021-05-04
use TransactionTestCase for asgi test

jkarstens:


easyaudit/tests/test_app/tests.py
```diff
@@ -24,14 +24,14 @@
...
+class WithUserInfoMixin:
+    def setUp(self):
+        self.username = 'joe@example.com'
+        self.email = 'joe@example.com'
+        self.password = 'password'
 
 
-@override_settings(TEST=True)
-class TestDjangoCompat(TestCase):
+class TestDjangoCompat(SimpleTestCase):
```
jkarstens, line 34: Didn't look like a database was needed for this test

jheld:
I have a feeling this did work in the past -- perhaps I broke it on a subsequent code change, or something else changed. How challenging will this be to keep the test passing while not changing the super-class?

easyaudit/tests/test_app/tests.py
jheld: Indeed that is the case. I think it's simpler to use the standard class but agreed, we're actually asserting django logic, not reaching into any db connection requirements.

jkarstens:
@jheld how can I help move this PR along? Looks like a workflow is awaiting approval. `TestASGIRequestTest` fails on Python >= 3.8.8 without `TransactionTestCase`

jheld:
@jkarstens do we know what difference in 3.8.8 caused the failure?

jkarstens:
@jheld thanks, appreciate the workflow approval. My guess is it's a change that runs sync code in a separate thread than the calling async code
https://github.com/soynatan/django-easy-audit/issues/182#issuecomment-831649014

The separate thread is documented in the Django docs (last sentence on this page https://docs.djangoproject.com/en/3.0/topics/async/), though I can't find the associated Python change (maybe it was an asgi change?)

jkarstens:
@jheld do you see any blockers for merging this?

PR #202, opened 2021-11-30
Fixed crash when auditing on binary data

boxed:


boxed:
Updated to handle the new django 4.x compatible code.

easyaudit/utils.py
```diff
@@ -19,15 +17,21 @@ def get_field_value(obj, field):
     :return: The value of the field as a string.
     :rtype: str
     """
+    raw_value = getattr(obj, field.name, None)
```
jheld, line 20: well placed!

easyaudit/utils.py
```diff
@@ -19,15 +17,21 @@ def get_field_value(obj, field):
...
         # to its naive form before we can accurately compare them for changes.
         try:
-            value = field.to_python(getattr(obj, field.name, None))
+            value = field.to_python(raw_value)
             if value is not None and settings.USE_TZ and not timezone.is_naive(value):
                 value = timezone.make_naive(value, timezone=timezone.utc)
         except ObjectDoesNotExist:
             value = field.default if field.default is not NOT_PROVIDED else None
+    elif isinstance(raw_value, bytes):
+        if len(raw_value) > 100:
```
jheld, line 31: I'm not _saying_ this is a magic number, but it would be helpful to have a comment regarding this (and possibly make it a variable since it is referenced multiple times in this function scope).

jheld (changes requested):
Overall looks great, thank you for working on this.

easyaudit/utils.py
boxed: It's being truncated because the audit log can't by default store a full copy of binary files that can be gigabytes in size.

mschoettle: Would it be worth (possible even?) storing a hash of the binary data instead of the first 100 bytes?

boxed: Hmm.. that's an idea for sure. The advantage is that you can actually check if some data is the exact data. The downsides are that you don't get something immediately useful in the log, and that if you don't have the binary data you want to compare to anymore then the hash is useless.

samamorgan: I would worry about the speed of hashing very large binary data. Say for a 1GB file, I'd expect at least 5 seconds for a hash to be generated if the implementation was purely C.

What about just a `len(bytes)` call?

easyaudit/utils.py
```diff
@@ -19,15 +17,21 @@ def get_field_value(obj, field):
...
+            return repr(raw_value[:100]) + '[truncated {} bytes]'.format(len(raw_value) - 100)
```
samamorgan, line 32:
```suggestion
            return repr(f"{raw_value[:100]}...[truncated {len(raw_value) - 100} bytes]")
```

PR #209, opened 2022-01-05
replace smart_text and force_text with smart_str and force_str

david-murray-wayflyer:
The _text versions have been deprecated since Django 3.0

https://docs.djangoproject.com/en/3.0/releases/3.0/#django-utils-encoding-force-text-and-smart-text

The _str versions work the same in all Python versions supported by
easyaudit (specifically, in Python 3).

Cleans up many of the deprecation warnings you will see from running
the easyaudit tests with `python -Wa`.


jheld:
Thank you for your time and effort but this was fixed by a different PR.

PR #217, opened 2022-02-19
fix: migration missing for field ID

### LINKED ISSUE #214: Migration missing for field ID
Migrations were not included for commit [fd05524dec14316a29191ad3e2934b99ee44843c](https://github.com/soynatan/django-easy-audit/commit/fd05524dec14316a29191ad3e2934b99ee44843c).

Please add the missing migrations so that running "makemigrations" is not required after installing django-easy-audit.

aymaneMx:
closes #214


mschoettle:
Thanks for taking care of creating a PR for this.

I think you could get around creating explicit `id` fields for each model by defining `DEFAULT_AUTO_FIELD` in the app config (see here for an example: https://docs.djangoproject.com/en/3.2/releases/3.2/#customizing-type-of-auto-created-primary-keys).

If I understand it right it would then not even require a new migration.

jheld:
I concur, the setting might be the optimal implementation. 

I was initially considering the possibility of if the parent project could make the decision (as opposed to the library app), and although I _do_ think that strategy should be supported by Django's migration constructs more easily, it is not likely a big deal for the library to make a choice.

Forks can be made, subclassed models can be made.

Can you make that change instead?

aymaneMx:
@jheld sure

Issue #190, opened 2021-06-22
delete possible despite not having the permission


nshmn:
Hi,
i'm creating an additional non-superuser admin site where the users should be able to view all easy audit entries but not change or delete them (so readonly). 
to do that i created a group that has only the view permission

`def add_site_admin_group(*args, **kwargs):
    site_admin_group, _created = Group.objects.get_or_create(name='site_admin')
    easyaudit_types = ContentType.objects.filter(app_label='easyaudit')
    easyaudit_permissions = Permission.objects.filter(content_type__in=easyaudit_types, codename__startswith='view')
    site_admin_group.permissions.set(easyaudit_permissions)
`
with this group i'm still able to delete entries in the admin panel (i'm successfully prevented from change though)

cybelew:
Looks like it's due to this section in EasyAuditModelAdmin:

https://github.com/soynatan/django-easy-audit/blob/93ab8d9a4b034cd4bde931897705b355e0c0c5a3/easyaudit/admin_helpers.py#L70-L73

Should probably read:

```python
    def has_delete_permission(self, request, obj=None):
        if settings.READONLY_EVENTS:
            return False
        return super().has_delete_permission(request, obj)

```

cybelew:
Additional note: This bug was introduced in v1.2.3

PR #224, opened 2022-03-14
Fix EasyAuditModelAdmin.has_delete_permission()

cybelew:
Fix for #190. EasyAuditModelAdmin.has_delete_permission() now respects user permissions if READONLY_EVENTS is False


jheld (approved):

jheld:
released to pypi https://pypi.org/project/django-easy-audit/1.3.3b2/

PR #232, opened 2022-06-06
Avoid DB savepoints if no audit is going to be created

hugobranquinho:
If we save a model with a db transaction and that model don't have audit active, it's making unnecessary db calls:
```
SAVEPOINT "xpto_x1";
RELEASE SAVEPOINT "xpto_x1";
```

This happens because django creates a save point if a transaction is open inside another transaction.
To avoid it, and save some db calls, do `should_audit(instance)` before creating `transaction.atomic`.

### DIFF
--- a/easyaudit/signals/model_signals.py
+++ b/easyaudit/signals/model_signals.py
@@ -54,9 +54,10 @@ def pre_save(sender, instance, raw, using, update_fields, **kwargs):
         return
 
     try:
+        if not should_audit(instance):
+            return False
+
         with transaction.atomic(using=using):
-            if not should_audit(instance):
-                return False
             try:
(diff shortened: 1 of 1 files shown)


jheld:
So I do agree that if we can avoid certain paths in the flow, we should, and I'm still concurring that this is a good change, but is it _actually_ hitting the database in the situations where the audit isn't going to be performed? Or is it more that we're looking to avoid needless extra (or any at all) db save points in the current transaction?

hugobranquinho:
@jheld 
Yup, is hitting the db to create and release the save point (2 db queries)
Because on my project we don't have audit for all tables (i think we have more without audit) every change we do on those, we have 2 extra queries. For example, if we make 50 object updates and none of those have audit, we have 150 db calls.

PR #233, opened 2022-06-06
Allow to ignore request.user validation

hugobranquinho:
If setting `DJANGO_EASY_AUDIT_CHECK_IF_REQUEST_USER_EXISTS` set to `False`, no `request.user` validation is made on DB.
This allow the calling project to ignore user validation and speed up audit creation, plus, save some DB queries.

### DIFF
--- a/easyaudit/signals/model_signals.py
+++ b/easyaudit/signals/model_signals.py
@@ -46,6 +46,23 @@ def should_audit(instance):
     return True
 
 
+def get_current_user_details():
+    user_id = None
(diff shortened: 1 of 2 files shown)


easyaudit/signals/model_signals.py
```diff
@@ -46,6 +46,23 @@ def should_audit(instance):
     return True
 
 
+def get_current_user_details():
```
jheld, line 49: If nothing else, thank you for the improved modularity!

PR #240, opened 2022-09-25
fix: fix potential slow query on huge requestevent table

### LINKED ISSUE #239: Poor performance due to datetime column is not indexed
Hi, we noticed very slow performance on showing the logs in production.
This is due to the default sorting on "datetime" column, but it's not indexed.

Normally people would not notice this issue unless they have massive logs like us.

Please add the index. Thx!

3cp:
closes #239


3cp:
A side layman question:
I am quite new to Django plugin development. In order to update the migration files, first I installed the project folder as a depenendy in my venv `pip3 install .`, then in the `easyaudit/tests/` folder, I run `python3 manage.py makemigrations easyaudit`. That created new migration file in the venv installed site-packages folder.

Then I copied out the new migration file back to the project folder like this:

    cp ../../venv/lib/python3.10/site-packages/easyaudit/migrations/0017_alter_requestevent_datetime.py ../migrations/

Is this the expected way to manage django plugin development? Is there a way to directly work on the files instead of jumping through the installed package in `venv/lib/pythonxxx/site-packages/`?

jheld:
Hi,

I usually install the project in editable mode `python -m pip install -e .` , which would at least get around the issue of where certain things live at all :)    but to your question about the migrations creation:

I _think_ I create migrations from the root but I specify the project settings folder e xplicitly (perhaps against the test projects settings). It has been awhile since I've created a migration PR but at least right now in my head that's what I would do. So, correct, no need to copy anything if running it via the way I have pseudo-written out.

jheld (approved):

jheld (approved):

Issue #235, opened 2022-07-20
user not recorded for requests when using a custom SESSION_ENGINE


oldmanstick:
I am building a system that uses a custom SESSION_ENGINE backed by a REST microservice instead of the standard database-backend session engine, and easy-audit does not record the user for requests.

Looks like the request_started_handler signal handler is using the Session model to look up the session instead of the app's configured SESSION_ENGINE. Is there a compelling reason for this?

jheld:
I'm unsure. Would you be able to post a WIP PR showing the requested change?

PR #242, opened 2022-12-02
Fix cookie parsing

### LINKED ISSUE #241: User object not recorded
I noticed that the User object was not being recorded for request events when users were authenticated with either django-rest-know or django-allauth. Is anyone else running into this?

steverecio:
Fixes #241


steverecio:
@soynatan Fixed the cookie parsing that was causing #241 . Might also be worth allowing users to configure a `get_current_user` function to account for different session engines as mentioned in #235

jheld:
> Still happening in 2026. This bug cost me hours of debugging after migrating to HTTPS. The issue is confirmed: SimpleCookie fails with modern JSON cookies (like Smartlook/GA4), while parse_cookie works. Can we finally merge this?

can you also add in support to default to the existing session engine as described in the final notes of https://github.com/soynatan/django-easy-audit/pull/242#issuecomment-1335810213 ? this _feels_ like a breaking change otherwise, and seems simple enough to give power back to the user and let everyone have their version of working code.

You may have to fork this branch in order to do so, but that's fine with me.

PR #249, opened 2023-03-14
Implemented CRUD_EVENT_NO_CHANGED_FIELDS_SKIP

felixcheruiyot:
I experienced issues with replicas when pre_save is used. To avoid this issues, I'd like to silence pre_saves.

I saw we have the flag CRUD_EVENT_NO_CHANGED_FIELDS_SKIP already in documentation but not implemented it. Please let me know if this fix implements CRUD_EVENT_NO_CHANGED_FIELDS_SKIP as it was intended.

Thanks


jheld:
Can you please the resolve the conflicts first?

PR #251, opened 2023-03-27
Fixes #250

### LINKED ISSUE #250: Add index for CRUDEvent.datetime
In our scenario we have 1.7M records in the table (postgres DB)

By indexing the field CRUDEvent.datetime the following query performed in the admin goes from 38seconds down to <0.5sec
```
SELECT "easyaudit_crudevent"."id",
       "easyaudit_crudevent"."event_type",
       "easyaudit_crudevent"."object_id",
       "easyaudit_crudevent"."content_type_id",
       "easyaudit_crudevent"."object_repr",
       "easyaudit_crudevent"."object_json_repr",
       "easyaudit_crudevent"."changed_fields",
       "easyaudit_crudevent"."user_id",
       "easyaudit_crudevent"."user_pk_as_string",
       "easyaudit_crudevent"."datetime"
  FROM "easyaudit_crudevent"
 ORDER BY "easyaudit_crudevent"."datetime" DESC, "easyaudit_crudevent"."id" DESC
 LIMIT 100
```

GigiusB:
Adding a descendent index on CRUDEvent.datetime in order to optimize retrieval times


jheld:
How bad would it be if this index were not descendant?

PR #254, opened 2023-05-11
Export audit events to CSV

Tim-Bolhoeve:
Added ability to export easyaudit events to CSV


Tim-Bolhoeve:
Ever since i implemented this, I have checked if there are styling changes on the Django admin panel that uses 3rd party styling like Django-Jazzmin. It seems to have no effect on the styling at all.

jheld (approved):

PR #264, opened 2023-11-22
chore: update versions for build workflow

mschoettle:
Update action versions, runner version and add Python 3.12 to the test matrix.


mschoettle:
Merged with #265

PR #266, opened 2024-01-03
raise exceptions only when they should be propagated

bernhardmiller:
since recent versions, easyaudit raises exceptions when something goes wrong in the signal handlers, independent of whether it is configured to do so or not.

the crud_flow functions are not called immediately, but passed as callbacks to be called later on. thus, the checked propagation in the exception handlers below is bypassed.


mschoettle:
@jheld Could you please release a new alpha version with this fix? We are facing an issue where the signal handler is raising an exception even though it is configured not to and I believe this error is fixed with this PR.

Issue #263, opened 2023-10-17
Make REMOTE_ADDR optional


mschoettle:
The `user_logged_in` signal handler fails during tests when calling `client.force_login(...)` because the request has no IP (`REMOTE_ADDR`).

```py
def user_logged_in(sender, request, user, **kwargs):
        try:
            with transaction.atomic(using=DATABASE_ALIAS):
                login_event = audit_logger.login({
                    'login_type': LoginEvent.LOGIN,
                    'username': getattr(user, user.USERNAME_FIELD),
                    'user_id': getattr(user, 'id', None),
>                   'remote_ip': request.META[REMOTE_ADDR_HEADER]
                })
E               KeyError: 'REMOTE_ADDR'

[snip]/easyaudit/signals/auth_signals.py:21: KeyError
```

Would it be ok to change this to `request.META.get(REMOTE_ADDR_HEADER)`? `remote_ip` is already nullable so from the `LoginEvent` model perspective it is already supported. And the request signal handler already supports this: 

https://github.com/soynatan/django-easy-audit/blob/bbdf834d671811043e7bab4aaeb39903ac21f982/easyaudit/signals/request_signals.py#L44

mschoettle:
@jheld Let me know what you think please. Happy to create a PR for this.

The alternative would be to disable auth events to be watched in tests but it would be nice to stay as close as possible to production.

jheld:
Should any extra conceptual weirdness be at play here by making this change, it's evading me at this time. Given the examples you've shown in the code where it certainly allows `None`, I concur with your approach. Yes, please make a PR :)

6abc:
# if we modify auth_signals.py 
# like this it will work, can u pls make these changes ?
from django.contrib.auth import signals, get_user_model
from django.db import transaction
from django.utils.module_loading import import_string

from easyaudit.middleware.easyaudit import get_current_request
from easyaudit.models import LoginEvent
from easyaudit.settings import REMOTE_ADDR_HEADER, WATCH_AUTH_EVENTS, LOGGING_BACKEND, \
    DATABASE_ALIAS
from easyaudit.utils import should_propagate_exceptions

audit_logger = import_string(LOGGING_BACKEND)()


def user_logged_in(sender, request, user, **kwargs):
    try:
        with transaction.atomic(using=DATABASE_ALIAS):
            login_event = audit_logger.login({
                'login_type': LoginEvent.LOGIN,
[...]

PR #267, opened 2024-01-05
Modernize package

samamorgan:
Modernizes this package to use the ideal tools for building, packaging, testing, and linting/formatting.

## Change summary

- Added Poetry as a requirement
- Added Ruff linting/formatting rules to ensure code quality
- Added djLint for Django template linting/formatting
- Added `pre-commit` as a requirement with git hooks for Ruff and djLint
- Converted tests from `unittest` to `pytest`
  - Moved tests out of package directory so they aren't included in distributed builds
- Added a build/publish action that automates package distribution
- Added a contributing guide
- Updated the test action to use new tools and include a code coverage comment for PRs (example at the bottom of [this commit](https://github.com/samamorgan/django-easy-audit/commit/865b44e2dedf751971e10311d40b5f2f41889564))
- Removed Travis configuration since this repo does not use it
- Removed code that resolved issues in older versions of Django and Python that this package does not support
- Refactored model signals to fix multiple complexity concerns

I've added extensive notes explaining the changes in code review comments. If anything is unclear, please ping me here!


pyproject.toml
```diff
@@ -0,0 +1,121 @@
...
+    "TID", # flake8-tidy-imports
+    "ERA", # eradicate
+    "PL",  # pylint
+    "RUF", # ruff
+]
+show-fixes = true
+target-version = "py38"
+
+[tool.ruff.lint.flake8-pytest-style]
+fixture-parentheses = false
```
samamorgan, line 117: These two just allow one to omit `()` from pytest fixture and mark decorators.

setup.py
samamorgan, line 1: No longer necessary, `pyproject.toml` takes care of all of this.

tests/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+import pytest
+from asgiref.sync import sync_to_async
+
+if TYPE_CHECKING:
+    from django.contrib.auth.models import User
+    from pytest_django.fixtures import SettingsWrapper
+
+
+@pytest.fixture(autouse=True)
+def test_settings(settings: SettingsWrapper) -> SettingsWrapper:
```
samamorgan, line 14: This adds `TEST = True` to Django settings for all tests.

tests/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+    settings.TEST = True
+
+    return settings
+
+    return settings
+
+
+@pytest.fixture
+def no_changed_fields_skip(settings: SettingsWrapper) -> SettingsWrapper:
```
samamorgan, line 23: Replaces `@override_settings`

tests/test_app/models.py
```diff
@@ -0,0 +1,53 @@
+# ruff: noqa: A003
+import uuid
+
+from django.db import models
+
+
+class Model(models.Model):
+    id = models.AutoField(primary_key=True)
```
samamorgan, line 8: Explicitly setting an ID here so Django checks don't get mad.

tests/test_app/models.py
samamorgan, line 1: Most of the changes here are just model name changes. Everything was `Test*` before, which makes pytest discovery angry (throws a bunch of warnings on every test run since pytest looks for classes named `Test*`.

tests/test_main.py
```diff
@@ -0,0 +1,371 @@
...
+
+@pytest.mark.django_db
+def test_no_migrations(capsys: pytest.CaptureFixture):
+    management.call_command("makemigrations", dry_run=True)
+
+    captured = capsys.readouterr().out
+    assert "No changes detected" in captured
+
+
+def test_no_issues(capsys: pytest.CaptureFixture):
```
samamorgan, line 34: Runs the Django system check framework to spot common problems. https://docs.djangoproject.com/en/5.0/topics/checks/

tests/test_main.py
```diff
@@ -0,0 +1,371 @@
...
+    management.call_command("check", fail_level="WARNING")
+
+    captured: str = capsys.readouterr().out
+    assert "System check identified no issues" in captured
+
+
+@pytest.mark.parametrize(
```
samamorgan, line 41: Parametrized this to run against every single model. IMO this test provides no value and should be deleted entirely. It is not this package's responsibility to test Django internals.

tests/test_main.py
```diff
@@ -0,0 +1,371 @@
...
+    def test_model_state(self, model):
+        """Ensures models have the internal `_state` object."""
+        model_instances = model()
+        assert hasattr(model_instances, "_state")
+
+
+@pytest.mark.django_db
+class TestAuditModels:
+    @pytest.fixture
+    def model(self):
```
samamorgan, line 65: The previous class attributes were moved to fixtures. This is the pytest way, and adds modularity since one can now add logic to these functions on subclasses if necessary.

tests/test_main.py
```diff
@@ -0,0 +1,371 @@
...
+    @pytest.fixture
+    def fk_model(self):
+        return ForeignKeyModel
+
+    @pytest.fixture
+    def m2m_model(self):
+        return M2MModel
+
+    @pytest.fixture
+    def _audit_logger(self, monkeypatch):
```
samamorgan, line 77: Using pytest monkey-patching instead of unittest mock.

tests/test_main.py
```diff
@@ -0,0 +1,371 @@
...
+            raise ValueError("Test exception")
+
+        monkeypatch.setattr("easyaudit.signals.crud_flows.audit_logger.crud", _crud)
+
+    def test_create_model(self, model):
+        obj = model.objects.create()
+        crud_event_qs = CRUDEvent.objects.filter(
+            object_id=obj.id, content_type=ContentType.objects.get_for_model(obj)
+        )
+        assert crud_event_qs.count() == 1
```
samamorgan, line 88: All of these now use `assert *` instead of the Django unittest style `assertFoo`.

tests/test_main.py
samamorgan, line 1: No measurable logical changes to the tests as they were previously written. I did my best to keep things identical when converting to `pytest` so the refactorings that were made in `easyaudit` could be proven as idempotent.

tests/test_main.py
```diff
@@ -0,0 +1,371 @@
...
+        assert data["fields"]["name"] == obj.name
+
+    def test_fk_model(self, model, fk_model):
+        obj = model.objects.create()
+        obj_fk = fk_model(name="test", test_fk=obj)
+        obj_fk.save()
+
+        crud_event = CRUDEvent.objects.filter(
+            object_id=obj_fk.id, content_type=ContentType.objects.get_for_model(obj_fk)
+        ).first()
```
samamorgan, line 101: Minor change here, this used to be `qs[0]`. `qs.first()` provides an identical result, but uses the built-in Django queryset function instead, which is more "correct".

tests/test_main.py
```diff
@@ -0,0 +1,371 @@
...
+        obj_m2m.test_m2m.clear()
+
+        crud_event = CRUDEvent.objects.filter(
+            object_id=obj_m2m.id,
+            content_type=ContentType.objects.get_for_model(obj_m2m),
+        ).first()
+        data = json.loads(crud_event.object_json_repr)[0]
+        assert [str(d) for d in data["fields"]["test_m2m"]] == []
+
+    @pytest.mark.usefixtures("no_changed_fields_skip")
```
samamorgan, line 132: Replaces `@override_settings`

tests/test_main.py
```diff
@@ -0,0 +1,371 @@
...
+        assert crud_event_qs.count() == 1
+
+        obj_id = obj.pk
+        obj.delete()
+        crud_event_qs = CRUDEvent.objects.filter(
+            object_id=obj_id, content_type=ContentType.objects.get_for_model(obj)
+        )
+        assert crud_event_qs.count() == 2
+
+    @pytest.mark.usefixtures("_audit_logger")
```
samamorgan, line 192: Replaces the previous `unittest.mock` usage.

tests/test_main.py
```diff
@@ -0,0 +1,371 @@
...
+        settings.DJANGO_EASY_AUDIT_PROPAGATE_EXCEPTIONS = False
+        try:
+            model.objects.create()
+        except ValueError:
+            pytest.fail("Unexpected ValueError")
+
+
+class TestAuditUUIDModels(TestAuditModels):
+    @pytest.fixture
+    def model(self):
```
samamorgan, line 206: Now we override the fixtures instead of the class attributes.

tests/test_main.py
```diff
@@ -0,0 +1,371 @@
...
+        return BigIntForeignKeyModel
+
+    @pytest.fixture
+    def m2m_model(self):
+        return BigIntM2MModel
+
+
+@pytest.mark.django_db
+class TestMiddleware:
+    def test_middleware_logged_in(self, user, client, username, password):
```
samamorgan, line 234: `pytest-django` provides a handy `client` fixture, so we no longer have to inherit from Django's `TestCase`: https://pytest-django.readthedocs.io/en/latest/helpers.html#client-django-test-client

tests/test_main.py
```diff
@@ -0,0 +1,371 @@
...
+        client.post(create_obj_url)
+        assert Model.objects.count() == 1
+
+        obj = Model.objects.all().first()
+        crud_event = CRUDEvent.objects.filter(
+            object_id=obj.id, content_type=ContentType.objects.get_for_model(obj)
+        ).first()
+        assert crud_event.user is None
+
+    def test_manual_set_user(self, django_user_model, username, email, password):
```
samamorgan, line 256: `django_user_model` is just a shortcut and avoids having to import `User`: https://pytest-django.readthedocs.io/en/latest/helpers.html#django-user-model

tests/test_main.py
```diff
@@ -0,0 +1,371 @@
...
+        assert Model.objects.count() == 1
+
+        obj = Model.objects.first()
+        crud_event = CRUDEvent.objects.filter(
+            object_id=obj.id, content_type=ContentType.objects.get_for_model(obj)
+        ).first()
+        assert crud_event.user == user
+
+
+@pytest.mark.asyncio
```
samamorgan, line 295: This now uses `pytest-asyncio`. Previously there was a `@tag` here, but now we use the asyncio mark to both mark these tests as async and retain the ability to only run async tests if desired.

tests/test_main.py
```diff
@@ -0,0 +1,371 @@
...
+@pytest.mark.django_db(transaction=True)
```
samamorgan, line 296: `transaction=True` makes this behave the same as if this class inhereted from `TransactionTestCase`

tests/test_main.py
```diff
@@ -0,0 +1,371 @@
...
+class TestASGIRequestEvent:
+    async def test_login(self, async_user, async_client, username, password):
```
samamorgan, line 298: https://pytest-django.readthedocs.io/en/latest/helpers.html#async-client-django-test-asyncclient

tests/test_main.py
```diff
@@ -0,0 +1,371 @@
...
+        await sync_to_async(async_client.login)(username=username, password=password)
+        assert await sync_to_async(RequestEvent.objects.count)() == 0
```
samamorgan, line 300: Note that all of the `sync_to_async` calls can be removed in favor of using Django's async query methods starting with Django 4.2. This package should drop support for Django <4.2 in April of this year to follow Django's cycle: https://www.djangoproject.com/download/#supported-versions (I'll add an issue for this)

easyaudit/signals/auth_signals.py
```diff
@@ -14,12 +18,14 @@
...
-                'username': getattr(user, user.USERNAME_FIELD),
-                'user_id': getattr(user, 'id', None),
-                'remote_ip': request.META[REMOTE_ADDR_HEADER]
-            })
+            audit_logger.login(
+                {
+                    "login_type": LoginEvent.LOGIN,
+                    "username": getattr(user, user.USERNAME_FIELD),
+                    "user_id": getattr(user, "id", None),
+                    "remote_ip": request.META.get(REMOTE_ADDR_HEADER) or "",
```
samamorgan, line 26: I was hitting issues with this when testing. This resolved #263

tests/test_main.py
```diff
@@ -0,0 +1,371 @@
...
+        assert RequestEvent.objects.count() == 0
+
+        resp = client.get(reverse("test_app:index"))
+        assert resp.status_code == 200
+
+        assert RequestEvent.objects.get(user=user)
+
+
+@pytest.mark.django_db
+class TestAuditAdmin:
```
samamorgan, line 355: Changes to this test removed the need for `BeautifulSoup`, it now uses Django's built-in `assertInHTML`. Also of note is that the HTML structure changed in Django 4.1 to use `summary` tags instead of `h3`

tests/test_app/migrations/0001_initial.py
samamorgan, line 1: I deleted all of the test app migrations and just made a new, fresh one. Should be no issue here, this is test code. The worst possible outcome is contributors will have to delete their local sqlite db, which is no real problem.

mschoettle:
Awesome work! 👍

samamorgan:
Thanks!

The test run may fail on creating a coverage comment. The code may have to be merged for that step to have the necessary permissions.

jheld:
I do have some concerns about the amount of changes taking place in a single PR. Many of these modifications have a reasonable backing for sure, and help to shed some of the baggage of years past, at the very least.

I am a bit hesitant though that all of it may fall under the umbrella of _ideal_, which implies subjectivity (this is software created by humans, subjectivity is all over the place, grain of salt applied). For instance, I'm not sure that I particularly want poetry in this project at this time. We would benefit from having a `pyproject.toml` , and `precommit`, and linting (like `ruff`) usage, absolutely, to name a few.

So I'm arguing can we split out some things and try to apply them to the project in series? Especially with such a large set of changed files from all of this.

I'm open to being persuaded of course.

samamorgan:
> I do have some concerns about the amount of changes taking place in a single PR. Many of these modifications have a reasonable backing for sure, and help to shed some of the baggage of years past, at the very least.

100% agree. This is a large PR, and well outside what I would usually consider a reasonable number of changes. To be fair, a majority of the changes were automatic, performed by Ruff linting. Since there was no linting standard on this package previously, it's not totally unexpected that the line count change is large given the number of contributors here and different code styles.

The only real logical changes you should see are in the refactoring of the crud flows, and even that was done in the name of reducing complexity according to ruff rules. I did this refactoring before swapping the test suite over to `pytest`, and the old test suite did pass against these changes, so I have a high degree of confidence that the changes are idempotent.

> I am a bit hesitant though to agree that all of it may fall under the umbrella of _ideal_, which implies subjectivity (this is software created by humans, subjectivity is all over the place, grain of salt applied). For instance, I'm not sure that I particularly want poetry in this project at this time. We would benefit from having a `pyproject.toml` , and `precommit`, and linting (like `ruff`) usage, absolutely, to name a few.

Ideal is certainly not the best word choice. The tools I've added here are generally the most common, active tools that many projects are currently using. I'm quite opinionated about all of this given my experience with these tools, and have benefited greatly from using Poetry especially to automate a lot of the manual things I had to keep in my head before. I particularly like Poetry because it solves a lot of the really strange, hard to resolve dependency issues I've had on larger projects in the past. Plus it comprehensively handles building, venv, etc. It's been pretty fantastic for me in the last ~2 years of use; it works well enough I've converted all of my personal projects and successfully argued for its use in my professional life on a quite large project. Overall the inclusion of that one tool has replaced several others and reduced the cognitive overhead across my entire team.

> So I'm arguing can we split out some things and try to apply them to the project in series? Especially with such a large set of changed files from all of this.

I'm all for this, but I'm concerned with the amount of work it will take. I actually considered doing this in stages at first, breaking each bulleted item I've listed above into a separate PR, but the individual PRs don't really illustrate the overall vision I have and I think it's harder to get some of this stuff past reviews without the entire context you see here.

My primary motivation here is to ease developement work for myself and, more importantly, newer contributors to `easyaudit`. When I started some work on a feature I had in mind, the ergonomics of development were very difficult and I'd really like to make things hopefully simpler for everyone.

samamorgan:
@jheld I changed the dependency groups to non-optional (optional wasn't necessary, I've been using that flag with incorrect assumptions):

https://python-poetry.org/docs/master/managing-dependencies/
> Dependency groups, other than the implicit main group, must only contain dependencies you need in your development process. Installing them is only possible by using Poetry.

I also made CI dispatchable to ease in testing workflow changes.

The next CI run should pass, with possibly the exception of the "Add coverage comment" step. I've seen this fail on changes from a fork, I believe because of this issue: https://github.com/actions/first-interaction/issues/10#issuecomment-562178406

mschoettle:
The one change I'd argue is helpful to extract is the one for #263 since it is unrelated to package organization, CI etc.

samamorgan:
> The one change I'd argue is helpful to extract into its own PR is the one for #263 since it is unrelated to package organization, CI etc.

I certainly could, but that change is required for the test suite to pass currently.

Let me know how or if you guys would like me to split this PR up. I'm happy to put the work in, just want to make some progress before this body of work gets out of my brain space.

mschoettle:
Yes that makes sense. Here is what I suggest: 

* Create a separate PR that only fixes #263 (if it was a separate commit it could be cherry-picked onto a separate branch but I could not see it only looking at the commit titles) [I am happy to do this as well]
* That PR gets merged first
* Then this PR here can be brought up to date and the tests will work
* Usually this PR here could be stacked on top of the other one to include the fix but it already does include it

I am just a user/contributor though :)

jheld:
concur it would be nice to have the remote addr PR in first.

samamorgan:
@jheld @mschoettle Good call, done: #272

CONTRIBUTING.md
```diff
@@ -0,0 +1,130 @@
...
+In order to determine whether you are dealing with a security issue, ask yourself these two questions:
+
+- Can I access something that's not mine, or something I shouldn't have access to?
+- Can I disable something for other people?
+
+If the answer to either of those two questions are _yes_, then you're probably dealing with a security issue. Note that even if you answer _no_ to both questions, you may still be dealing with a security issue, so if you're unsure just send an email.
+
+### Bugs
+
+When filing an issue, make sure to answer these five questions:
```
mschoettle, line 68: We could add an issue template for this in a future PR

easyaudit/signals/auth_signals.py
mschoettle: Moved into its own PR: #272

pyproject.toml
```diff
@@ -0,0 +1,115 @@
...
+    "Programming Language :: Python :: 3.12",
+    "Topic :: Software Development :: Libraries :: Python Modules",
+]
+packages = [{include = "easyaudit"}]
+
+[tool.poetry.dependencies]
+python = "^3.8"
+django = [
+    {version = ">=3.2,<5", python = "<3.10"},
+    {version = "~5", python = ">=3.10"},
```
mschoettle, line 37: Does this restrict it to Django 3.2 to 5? Or would a v6 work with this? I am unfamiliar with the poetry syntax. Asking because upper version bounds are problematic: https://iscinumpy.dev/post/bound-version-constraints/

pyproject.toml
```diff
@@ -0,0 +1,121 @@
...
+]
+
+[tool.ruff]
+extend-exclude = ["migrations"]
+ignore = [
+    "D1",     # Missing docstrings
+    "D203",   # Docstrings on class definitions not preceded by a blank line
+    "D213",   # Multi-line docstring summary should start at the second line
+    "D407",   # Dashed underline after doc section (not compatible with google style)
+    "DJ008",  # Model does not define `__str__` method
```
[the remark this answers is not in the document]
mschoettle: This could be done in a follow-up PR

pyproject.toml
```diff
@@ -0,0 +1,121 @@
...
+    "RUF012", # Mutable default values in class attributes
+]
+line-length = 92
+select = [ # https://docs.astral.sh/ruff/rules
```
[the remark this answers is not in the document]
mschoettle: Out of curiosity: Are there specific ones that you chose not to include?

easyaudit/migrations/0018_auto_20240105_0125.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('easyaudit', '0017_alter_requestevent_datetime'),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
mschoettle, line 13: What are all the `AlterField` operations caused by?

mschoettle (commented):
I went through the changes and added a few comments. Great work! I haven't used poetry myself before so not entirely sure what the benefit of it is. I rally hope that we can get this merged in.

One general comment: It seems that there is an inconsistent use of single and double quotes right now. Cam `ruff-format` take care of this? Otherwise, we should include the equivalent rule of `flake8-quotes`. Personally, I am in favour of using single quotes but fine either way.

CONTRIBUTING.md
samamorgan: Agreed, that would be a good call.

pyproject.toml
samamorgan: I think you're correct, this isn't quite right. This would restrict Django to `>=5.0.0 <6.0.0`. Here's the docs:

https://python-poetry.org/docs/dependency-specification/#tilde-requirements

I was trying to match the Python version to the supported Django version table from here:

https://docs.djangoproject.com/en/5.0/faq/install/#what-python-version-can-i-use-with-django

With that in mind, I think this should likely be `{version = ">=3.2.9,<6", python = ">=3.10"}`. I'd have to test that, but I think it would work correctly. Thoughts?

pyproject.toml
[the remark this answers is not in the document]
samamorgan: I don't think there are any I've explicitly decided to not include (with the exception of some explicit ignores defined in `ignore`). The list here is from another long-term project I've been involved in. There are a ton of ruff rules:

https://docs.astral.sh/ruff/rules/

easyaudit/models.py
```diff
@@ -18,82 +18,123 @@ class CRUDEvent(models.Model):
...
+        help_text=_("String version of the user pk"),
+        verbose_name=_("User PK as string"),
+    )
+    datetime = models.DateTimeField(auto_now_add=True, verbose_name=_("Date time"))
+
+    class Meta:
+        verbose_name = _("CRUD event")
+        verbose_name_plural = _("CRUD events")
+        ordering = ["-datetime"]
+        indexes = [models.Index(fields=["object_id", "content_type"])]
```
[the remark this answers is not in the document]
samamorgan: I don't think it's a necessary change, and I agree that #258 should likely merge first, since this is duplicated work. I'll just fix the merge conflicts after those two merge.

easyaudit/migrations/0018_auto_20240105_0125.py
samamorgan: Many of these fields allowed `null=True, blank=True`, which is bad. That means there are two possible values with the same meaning. This rule explains it well:

https://docs.astral.sh/ruff/rules/django-nullable-model-string-field/

I changed all of them to `blank=True, default=""`, which will not propagate `None` as a value, but still allow the field to be blank.

samamorgan:
> One general comment: It seems that there is an inconsistent use of single and double quotes right now. Cam `ruff-format` take care of this? Otherwise, we should include the equivalent rule of `flake8-quotes`. Personally, I am in favour of using single quotes but fine either way.

This is already being done:

https://docs.astral.sh/ruff/rules/bad-quotes-inline-string/

The inconsistency you're seeing is when there's a string that also contains double quotes, like `f'a = "b"``

Double quotes are generally better IMO, mostly because when you're trying to generate an HTML string like `f'<a href="{url}">{escaped_obj_repr}</a>'`, but also it's been kind of the standard in Python for a lot of years.

mschoettle:
> > One general comment: It seems that there is an inconsistent use of single and double quotes right now. Cam `ruff-format` take care of this? Otherwise, we should include the equivalent rule of `flake8-quotes`. Personally, I am in favour of using single quotes but fine either way.
> 
> Double quotes have been the default in `black` for as long as I've used it, so they seem to have become the de-facto standard in Python. It's such a standard preference that I'd actually have to add additional configuration to use single quotes instead:
> 
> https://docs.astral.sh/ruff/settings/#lint_flake8-quotes_inline-quotes
> 
> The rule you mentioned is already applied:
> 
> https://docs.astral.sh/ruff/rules/bad-quotes-inline-string/
> 
> The inconsistency you're seeing is when there's a string that also contains double quotes, like `f'a = "b"'`, which is automatically fixed by this rule:
> 
> https://docs.astral.sh/ruff/rules/avoidable-escaped-quote/

My bad, I missed the double (as in multiple) quotes. Thanks for the references.

pyproject.toml
mschoettle: I propose to only do `>=3.2.9"` without an upper bound. This allows users to test/use this package with a later Django version if they want (without having to wait for an update of this package). See the article I linked above.

CONTRIBUTING.md
```diff
@@ -0,0 +1,130 @@
...
+- Bug fixes that change default return values or error codes stored in constants
+- Adding logging messages or debugging output
+- Changes to ‘metadata’ files like .gitignore, build scripts, etc.
+- Moving source files from one directory or package to another
+
+## Reporting bugs
+
+### Security vulnerabilities
+
+If you find a security vulnerability, **DO NOT** open an issue. Email [natancalzolari@gmail.com](mailto:natancalzolari@gmail.com) instead so as to not expose the vulnerability to the public.
```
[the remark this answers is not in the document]
jheld: I think that's generally reasonable. I'm happy to be in correspondence but yes, there are some items where the nature of it may be more for @soynatan , and I would say so if that was my feeling. But in the end, also up to him :)

samamorgan:
@mschoettle @jheld Where do we stand on this one? This has been up for review for about 2 months now. I'd like to get a plan at least to get it merged so I don't have to keep fixing merge conflicts.

mschoettle:
I'm in favour of moving ahead with this. My suggestion would be to merge #276, #277 and then #258 before this one.

samamorgan:
> sorry @samamorgan , with all of the other PRs, we have a merge conflict.

@jheld What merge conflict are you seeing? I resolved what I saw this morning, and don't see anything else blocking according to GH.
image

jheld:
Unsure yet, just what github is saying

image

samamorgan:
I'd recommend changing the merge strategy to `squash and merge` if you can. Rebasing on every merge seems inappropriate.

samamorgan:
@jheld I pulled all of my work onto a separate branch, squashed all commits, rebased on master, then force-pushed all of this to my master branch. LMK if that resolves the issue.

jheld:
@samamorgan much appreciated. I hadn't inspected the merge option at that time so I it didn't occur to me that it might not have been a true conflict. That said, it is quite helpful for these changes to be atomic when possible.

jheld:
> I'm in favour of moving ahead with this. My suggestion would be to merge #276, #277 and then #258 before this one.

I had merged this and just reverted because I remember that #258 was not yet merged. If we don't need to wait for it, let me know and I'll make it right.

samamorgan:
> I had merged this and just reverted because I remember that #258 was not yet merged. If we don't need to wait for it, let me know and I'll make it right.

This PR does actually cover the changes that #258 covers, but if a clear change path is desired, it would technically be better to merge #258 first. @mschoettle That PR has test failures that need to be resolved.

@jheld How do we resolve this now that it's closed and has been reverted? Haven't quite been in this particular situation before.

jheld:
For keeping things transparent/open, I would have you open another PR with your changes. I _could_ attempt to do it behind the scenes locally once #258 is merged (which is the preference, first), but again, chain of changes tends to flow a little cleaner when going via the MR/PR route.

Apologies for this last hiccup.

mschoettle:
I’ll fix #258 ASAP to prevent blocking your PR

PR #271, opened 2024-01-30
Modify request_started_handler to use SESSION_ENGINE

oldmanstick:


easyaudit/signals/request_signals.py
```diff
@@ -1,11 +1,15 @@
...
+
+from django.contrib.auth import get_user_model, SESSION_KEY as AUTH_SESSION_KEY
 from django.contrib.sessions.models import Session
 from django.core.signals import request_started
 from django.http.cookie import SimpleCookie
 from django.utils import timezone
 from django.conf import settings
 from django.utils.module_loading import import_string
 
+session_engine = import_module(settings.SESSION_ENGINE)
```
jheld, line 11: I'm not immediately aware of any, but could there be any bugs with `import_module` being run at the global module level? For instance, in testing. So far my only hesitation with the changes.

jheld (commented):
just one open comment, otherwise this looks good!

easyaudit/signals/request_signals.py
oldmanstick: This is the method indicated in the Django documentation under “using sessions outside of views”.
[](https://docs.djangoproject.com/en/5.0/topics/http/sessions/#using-sessions-out-of-views)

Can you speak more about your concerns involving testing?

PR #276, opened 2024-03-14
fix: make REMOTE_ADDR optional in all cases

mschoettle:
Follow-up to #272. Make it optional everywhere, there were a few cases missing.

Refs #263


samamorgan (approved):
LGTM! 🚀

PR #277, opened 2024-03-14
Drop support for Django 3.2

### LINKED ISSUE #269: Drop support for Django <4.2
As of April 2024, Django will end support for v3.2-LTS. This package drop support for anything below Django v4.2-LTS in April to ensure new features aren't developed against insecure code.

This has the added benefit of being able to remove `sync_to_async` calls in tests, which can instead call queryset async methods (`aget`, `acount`, etc).

mschoettle:
Changes minimum supported version to Django 4.2 in preparation of LTS for Django 3.2 ending in April.

Fixes #269

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,8 +27,6 @@ setup(
     classifiers=[
         'Environment :: Plugins',
         'Framework :: Django',
-        "Framework :: Django :: 4.0",
-        "Framework :: Django :: 4.1",
         "Framework :: Django :: 4.2",
         "Framework :: Django :: 5.0",
         'Intended Audience :: Developers',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -27,7 +27,6 @@
     classifiers=[
         'Environment :: Plugins',
         'Framework :: Django',
-        "Framework :: Django :: 3.2",
```
samamorgan, line 30: Shouldn't you drop the 4.0 and 4.1 classifiers too?

mschoettle: Good catch! Yes, removed them.

samamorgan (approved):
🚀

PR #281, opened 2024-03-19
Fix semver

samamorgan:
Semantic versions must be in the form `MAJOR.MINOR.PATCH-PRERELEASE`. The currently applied semver violates the spec.

This is important for CD to function properly. An invalid semver string will cause the job to fail, as it did here: https://github.com/soynatan/django-easy-audit/actions/runs/8333068101/job/22803668383#step:7:9

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "django-easy-audit"
-version = "1.3.7.a1"
+version = "1.3.7-a1"
 description = "Yet another Django audit log app, hopefully the simplest one."
 license = "GPL3"
 authors = ["Natán Calzolari <natancalzolari@gmail.com>"]


jheld:
Makes sense. Unfortunately these rules are different than what python allowed, otherwise I would have used a different scheme a long time ago.

mschoettle:
So what is the solution to satisfy both semver and Python packaging (there it seems to require `major.minor.patch[.something]`?

jheld:
No i think we should be able to make the change. BTW the 1.3.7.a1 release is on PyPI.

samamorgan:
@jheld @mschoettle Hold up, will normal semver rules not work for publishing PyPI packages? If that's the case, I have some more work to do in `cd.yml` to make that work properly.

samamorgan:
If I'm reading the docs correctly, a dash should work fine. Oddly, the normal form is no separator, which I didn't expect.

https://packaging.python.org/en/latest/specifications/version-specifiers/#pre-release-separators
> Pre-releases should allow a ., -, or _ separator between the release segment and the pre-release segment. The normal form for this is without a separator. This allows versions such as 1.1.a1 or 1.1-a1 which would be normalized to 1.1a1.

So, reading that, I would expect a publish action with `x.y.z-a1` to go live to PyPI with `x.y.za1`

jheld:
Regarding the last point, I manually released to PyPI of course, but yes, even the URL/version listed don't match what I set the version as, how fun (it's trivial in this case, but def makes it not an identity function).

Issue #291, opened 2024-04-20
Upgrade `EasyAuditMiddleware`  to support sync and async contexts.


KaczuH:
`EasyAuditMiddleware` currently only supports synchronous code, requiring adaptations by Django for async operation.

As such, projects using easy audit cannot benefit from async stack as shown by the `django.request` logger:

`DEBUG:django.request:Asynchronous handler adapted for middleware easyaudit.middleware.easyaudit.EasyAuditMiddleware.`

Relevant Django documentation can be found in [Django Docs:](https://docs.djangoproject.com/en/5.0/topics/async/#async-views).

There is a potential simple fix: replacing `threading.local` with `asgiref` and using `MiddlewareMixin`.

reckless-huang:
This is a good PR, why is there no response?

jheld:
1.3.8-a1 has support for this. available on PyPI. I'm sure there will be more pre-releases before 1.3.8 is official, but still the feedback from it being used in the wild is very helpful.

PR #292, opened 2024-04-20
Update `EasyAuditMiddleware` to support async context

KaczuH:
Replaced standard threading with 'asgiref.local' in EasyAuditMiddleware. Also, made EasyAuditMiddleware extend Django's MiddlewareMixin to automatically handle sync and async execution modes. Github issue: #291

### DIFF
--- a/easyaudit/middleware/easyaudit.py
+++ b/easyaudit/middleware/easyaudit.py
@@ -1,6 +1,11 @@
 # makes easy-audit thread-safe
 import contextlib
-from threading import local
+from typing import Callable
+
+from asgiref.local import Local
+from asgiref.sync import iscoroutinefunction, markcoroutinefunction
+from django.http.request import HttpRequest
+from django.http.response import HttpResponse
 
 
 class MockRequest:
@@ -10,7 +15,7 @@ class MockRequest:
         super().__init__(*args, **kwargs)
 
 
-_thread_locals = local()
+_thread_locals = Local()
 
 
 def get_current_request():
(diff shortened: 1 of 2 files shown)


easyaudit/middleware/easyaudit.py
```diff
@@ -37,12 +38,9 @@ def clear_request():
         del _thread_locals.request
 
 
-class EasyAuditMiddleware:
+class EasyAuditMiddleware(MiddlewareMixin):
```
samamorgan, line 41: You should not convert this to an old-style mixin.

samamorgan (changes requested):

easyaudit/middleware/easyaudit.py
KaczuH: `EasyAuditMiddleware` already uses old pre-Django-1.10 convention. `MiddlewareMixin` helps achieve sync/async capability without much boilerplate code.

Would you be ok with going fully with new style middleware?

```suggestion
class EasyAuditMiddleware:
    """Makes request available to this app signals."""
    async_capable = True
    sync_capable = True

    def __init__(self, get_response):
        self.get_response = get_response
        if iscoroutinefunction(self.get_response):
            markcoroutinefunction(self)

    def __call__(self, request):
        if iscoroutinefunction(self):
            return self.__acall__(request)

        _thread_locals.request = request
        response = self.get_response(request)

        with contextlib.suppress(AttributeError):
            del _thread_locals.request

        return response

    async def __acall__(self, request):
        _thread_locals.request = request

        response = await self.get_response(request)

        with contextlib.suppress(AttributeError):
            del _thread_locals.request

        return response
```

samamorgan: Yep, that'd be ideal IMO. Get rid of legacy conventions.

Also, don't forget tests!

KaczuH: @samamorgan I've pushed requested changes.

jheld (approved):
As far I can tell, this is a good modernization effort.

jheld:
@KaczuH can you please squash these commits before I merge?

KaczuH:
@jheld Done!

PR #299, opened 2024-07-25
Use json fields where applicable

### LINKED ISSUE #298: Use JSONField for object repr field
Is there any reason why we can't use JSONField instead of TextField for object_json_repr? It would allow for querying within the json for specific logs.

https://docs.djangoproject.com/en/5.0/ref/models/fields/#jsonfield

sgordon16:
Resolves #298


jheld:
How well does the code work when querying rows after the migration goes through?

sgordon16:
I'm not sure I understand your question, but you can now do a query like this for example:
```
CRUDEvent.objects.filter(object_json_repr__0__model="app.user", changed_fields__has_key="name")
```
This will give you all the records of users that had their name changed.

sgordon16:
@jheld Just following up on this. Are there any issues with this?

jheld:
_Probably_ not. but we are in the middle of a potential issue on the latest pre-release and I'm hesitate to add more big changes like this. I think `1.3.8` series would be a good fit.

PR #304, opened 2024-08-20
Fix migrating db that isn't past squashed migration

PaarthShah:
Retroactively changing the migration is the only way for this to work: it's a safe operation on its own because:
1. The migration has already been run
2. the operation itself is identical (it was copied from the squashed migration)
3. for a fresh migration, only the squashed version is run anyway

Fixes #303

Migration state after:
```
easyaudit                                                                                                                                                                        
 [X] 0001_initial                                                                                                                                                                
 [X] 0002_auto_20170125_0759                                                                                                                                                     
 [X] 0003_auto_20170228_1505                                                                                                                                                     
 [X] 0004_auto_20170620_1354_squashed_0019_alter_crudevent_changed_fields_and_more (16 squashed migrations)
```

Issue #303, opened 2024-08-20
1.3.7b3 migration fails if only partially migrated


PaarthShah:
Migration status:
```
easyaudit                                                                                                                                                                                                          
 [X] 0001_initial                                                                                                                                                                                                  
 [X] 0002_auto_20170125_0759                                                                                                                                                                                       
 [X] 0003_auto_20170228_1505                                                                                                                                                                                       
 [X] 0004_auto_20170620_1354                                                                                                                                                                                       
 [X] 0005_auto_20170713_1155                                                                                                                                                                                       
 [X] 0006_auto_20171018_1242
 [X] 0007_auto_20180105_0838
 [X] 0008_auto_20180220_1908
 [X] 0009_auto_20180314_2225
 [X] 0010_repr_text
 [X] 0011_auto_20181101_1339
 [X] 0012_auto_20181018_0012
 [X] 0013_auto_20190723_0126
 [X] 0014_auto_20200513_0008
 [X] 0015_auto_20201019_1217
 [X] 0016_alter_crudevent_event_type
 [X] 0017_alter_requestevent_datetime
 [ ] 0018_rename_crudevent_object_id_content_type_index
 [ ] 0019_alter_crudevent_changed_fields_and_more
```

Error observed:
```
Operations to perform:                                                                                                                                                                                             
  Apply all migrations: easyaudit                                              
Traceback (most recent call last):                                                                                                                                                                                 
  File "/var/www/manage.py", line 22, in <module>                                                                                                                                                                  
    execute_from_command_line(sys.argv)                                                                                                                                                                            
  File "/usr/local/lib/python3.11/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line                                                                                        
    utility.execute()                                                                                                                                                                                              
  File "/usr/local/lib/python3.11/site-packages/django/core/management/__init__.py", line 436, in execute                                                                                                          
    self.fetch_command(subcommand).run_from_argv(self.argv)                                                                                                                                                        
  File "/usr/local/lib/python3.11/site-packages/django/core/management/base.py", line 413, in run_from_argv                                                                                                        
    self.execute(*args, **cmd_options)                                                                                                                                                                             
  File "/usr/local/lib/python3.11/site-packages/django/core/management/base.py", line 459, in execute                                                                                                              
    output = self.handle(*args, **options)                                                                                                                                                                         
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                         
[...]

Issue #301, opened 2024-08-12
Any plans to release v1.3.7 with support for Django 5.1?


scribejourdan:
I'm trying to upgrade Django to 5.1 but I'm unable because this package triggers an `indexed_together` exception on ver sion 1.3.6.

Additionally, on version 1.3.7b1, poetry throws this error:

```
Because django-easy-audit (1.3.7b1) depends on django (>=4.2,<5.0)
 and no versions of django-easy-audit match >1.3.7b1,<1.4.0, django-easy-audit (>=1.3.7b1,<1.4.0) requires django (>=4.2,<5.0).
So, because non-package-mode depends on both django-easy-audit (~1.3.7b1) and Django (~5.1), version solving failed.
```

jheld:
Great question. I've just released 1.3.7.b2 which intends to broaden the django support from >=4.2, <6.0 and I am hoping helps get around the issue you were seeing.

djch:
FWIW, I'm still getting the `index_together` error after installing the 1.3.7b3 release. 

No other instances of `index_together` in my codebase.

jheld:
Which django version are you on now?

I might not be able to check into this more short term but perhaps @mschoettle has any ideas?

djch:
On Django 5.1. 

If I downgrade to back 5.0.8, there's no problem (same as v1.3.6).

mschoettle:
@djch can you please provide the full stack trace of the error you are getting?

scribejourdan:
@mschoettle @jheld I'm having no issues anymore, thanks!

Despite of that, I'd really like to know if there's any projection on when 1.3.7 will be released i.e. not beta. It's a little hard to pitch an update based on beta versions.

scribejourdan:
Well, actually, I just faced the same issue @djch experienced.

Tracked it down to the migration [0004](https://github.com/soynatan/django-easy-audit/blob/master/easyaudit/migrations/0004_auto_20170620_1354.py#L16).

I saw a squashed 0004 migration there which should replace the original 0004, but Django still loads that migration, it seems. I placed a debugger there to make sure, it got hit.

I'm being able to run everything that don't trigger that migration, so I can start the server and run manage commands. This is happening solely when I run tests, I assume because the database gets built from scratch, so it goes through all migrations.

If you kept the original migrations concerned they'd be needed, you don't need to worry. As long as the squash migration reaches the exact same state as the other combined and they're referenced at the `replaces` property, it'll be alright.

[...]

PR #305, opened 2024-08-27
Remove squashed migrations as they're not guaranteed to be ignored

jourdanrodrigues:
Partially address #301, addresses #303 and replaces #304


jourdanrodrigues:
Closing in favor of #304

PR #306, opened 2024-08-29
ci: add django versions to CI matrix

mschoettle:
Add supported Django versions to CI matrix again.


mschoettle:
@jheld Would be great if we could merge #283 and then this one to ensure that we run tests with all supported Python and Django combinations.

mschoettle:
@jheld Thanks for merging #283. Finalized and rebased.

PR #323, opened 2025-01-15
Use renamed gettextlaxy for Django 4 compatibility

### LINKED ISSUE #322: CRUDHistoryAdminMixin does not support Django 4
1)  What version of Python and Django are you using?

Python: 3.12.3
Django: 4.2.11

2) What operating system and processor architecture are you using?
Ubuntu 24.04.1 LTS
x86_64

3) What did you do?
Import the `CRUDHistoryAdminMixin` to use in a ModelAdmin class.

```python
from easyaudit.crudhistory_admin_mixin import CRUDHistoryAdminMixin
```

4) What did you expect to see?
An import without errors.

5) What did you see instead?

Importing the Mixin is raising an ImportError because it uses `ugettext_lazy` which was [removed in Django 4.0](https://docs.djangoproject.com/en/dev/internals/deprecation/#deprecation-removed-in-4-0), replaced with `gettext_lazy`.

To replicate the issue in this project, if the above import is added to `tests/test_main.py`, pytest will output:
```
_____________________ ERROR collecting tests/test_main.py ______________________
ImportError while importing test module '/home/jt/code/django-easy-audit/tests/test_main.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_main.py:13: in <module>
    from easyaudit.crudhistory_admin_mixin import CRUDHistoryAdminMixin
easyaudit/crudhistory_admin_mixin.py:11: in <module>
    from django.utils.translation import ugettext_lazy as _
E   ImportError: cannot import name 'ugettext_lazy' from 'django.utils.translation' (/home/jt/code/django-easy-audit/.venv/lib/python3.12/site-packages/django/utils/translation/__init__.py)
```

The fix is as simple as updating the function name in [this line](https://github.com/soynatan/django-easy-audit/blob/d2941373dd2107bc7994884cb044fcd2af2bb086/easyaudit/crudhistory_admin_mixin.py#L11)  to `gettext_lazy`.

For reference, [this PR](https://github.com/soynatan/django-easy-audit/pull/193) added the feature.

hydrospanner:
Fixes #322

### DIFF
--- a/easyaudit/crudhistory_admin_mixin.py
+++ b/easyaudit/crudhistory_admin_mixin.py
@@ -8,7 +8,7 @@ from django.http import HttpRequest, HttpResponseRedirect
 from django.shortcuts import redirect
 from django.urls import path, reverse
 from django.utils.html import format_html
-from django.utils.translation import ugettext_lazy as _
+from django.utils.translation import gettext_lazy as _
 
 
 class BaseProcessActionsAdminMixin:


jheld:
I'm actually really surprised that no one has encountered this problem, yet. Maybe it's just because it's only for the admin.

hydrospanner:
> I'm actually really surprised that no one has encountered this problem, yet. Maybe it's just because it's only for the admin.

The exception was only raised when using the admin mixin. It's not a documented feature, so I assume few  use it.

PR #330, opened 2025-08-05
Tarasinf/ccs 2252 add new fileds into crudevent

tarasinf:
https://plentific.atlassian.net/browse/CCS-2252


tarasinf:
Sorry, wrong target.

PR #333, opened 2025-12-06
feat: add support for Django 6.0 and Python 3.14

ma11011s:
This pull request adds support for Django 6.0 and Python 3.14.

Tested the package with:
- Django 6.0
- Python 3.12, 3.13, and 3.14
- Middleware and model audit events
- Admin integration
- Migration compatibility


jberends:
merge it and patch release. See #334 

We have problems/annoyances/workarounds in production due to this limit here.

PR #340, opened 2026-03-11
Django upgrades, UUID handling

pcraciunoiu:
Upstreaming more changes from our fork that we've been using in production.

* Upgrade various imports/calls to use Django 4.2+ (used django-upgrade lib)
* Handle UUID in changed_fields, was giving issues
* Handle async transactions properly

Thanks for reviewing!


easyaudit/signals/crud_flows.py
```diff
@@ -97,10 +97,12 @@ def m2m_changed_crud_flow(  # noqa: PLR0913
         if action == "post_clear":
             changed_fields = []
         else:
-            changed_fields = json.dumps(
-                {get_m2m_field_name(model, instance): list(pk_set)},
-                cls=DjangoJSONEncoder,
-            )
+            pks = pk_set
+            if isinstance(pks, Iterable):
+                pks = (format_primary_key(pk) for pk in pks)
```
jheld: Maybe just make it a `list` here so we don't have to worry about converting it later.

easyaudit/signals/crud_flows.py
```diff
@@ -97,10 +97,12 @@ def m2m_changed_crud_flow(  # noqa: PLR0913
...
+            else:
+                pks = format_primary_key(pks)
```
jheld: if not an iterable, then what exactly is happening here? it's also been a long time since i've read the flow of how `pk_set` can be passed in.

jheld (commented):
i'm probably going to approve this anyway but hoping to get more information about the changes in crud_flows before doing it.

easyaudit/signals/crud_flows.py
pcraciunoiu: Good call, I'll simplify this and make sure we're testing it in one project too.

Pushed changes.

jheld: @pcraciunoiu have you been able to test out this change in one of your projects?

pcraciunoiu: @jheld it looks like I didn't get to creating a new tag on our fork and updating our project to use the changes after the last commit https://github.com/soynatan/django-easy-audit/pull/340/changes/bfe2ced4a05d2151a8bd7acdb1f2b89d681b638d

Otherwise it's been tested. I'll push that update and deploy, but it'll likely be a few days of testing to be sure (it's not a very active project)

jheld: I'm in no rush but I also haven't used this project in a company since 2020. If you're okay to wait on getting some runtime signals, that's stellar! If you're otherwise feeling confident, then we can merge and make a new release.

pcraciunoiu: Been using it for a few days. The main use case for the app is an audit trail of changes. No issues with these changes that I could see.

I do have a couple more features that I'd like to PR. The main one is to add a metadata with a callback so that using this project allows us to add custom data on top of the change events.

I'll PR those changes separately soon.

pcraciunoiu: @jheld meaning, I feel good about merging this if you are ready

pcraciunoiu: Opened follow-ups on top of this in #343 and #344.

Those last two bring parity with our fork, which would be ideal, but they are more substantial as well.

PR #344, opened 2026-06-05
Add metadata, custom queryset hooks, and field resolvers, m2m changes

pcraciunoiu:
This is based on #340 and #343.

Changes are only in 8b86281f6c4fac6b41fe27390da54e52636d9f8e and 1bf2f3fff548b7cd6a2ef795d5c6fb044e825a53

What follows is an AI generated summary.

## Summary

Extends django-easy-audit with optional model hooks for richer audit data and more flexible field/update handling:

- **`CRUDEvent.metadata`** — JSON field populated from per-model hooks (e.g. link related records for cross-model queries)
- **Custom audit queryset** — use a model-specific queryset when loading the pre-update instance (e.g. soft-delete / `all_with_deleted()`)
- **Pluggable field value resolvers** — customize how field values are compared in change deltas
- **`SimpleCookie` import fix** — use `http.cookies` instead of removed `django.http.cookie` (Django 6 compatibility)
- **M2M multi-table inheritance fix** — resolve M2M field names when the relation is declared on a parent model

All features are **opt-in** via model methods or settings, except the M2M inheritance fix which applies automatically. Models without hooks behave as before (`metadata` is `null`, default queryset and field handling unchanged).

## Changes

| Area | Files |
|------|-------|
| Metadata field + population | `easyaudit/models.py`, `easyaudit/signals/crud_flows.py`, `easyaudit/utils.py`, `easyaudit/migrations/0021_crudevent_metadata.py` |
| Custom queryset for update deltas | `easyaudit/utils.py`, `easyaudit/signals/model_signals.py` |
| Field value resolver map | `easyaudit/utils.py` |
| M2M multi-table inheritance | `easyaudit/utils.py` (`get_m2m_field_name`) |
| SimpleCookie import | `easyaudit/signals/request_signals.py` |
| Documentation | `README.md` |
| Tests | `tests/test_app/models.py`, `tests/test_app/migrations/0003_metadata_models.py`, `tests/test_app/migrations/0004_place_restaurant_m2m_inheritance.py`, `tests/test_main.py`, `tests/test_m2m_logging.py` |

### 1. Instance metadata (`CRUDEvent.metadata`)

Add a `metadata` JSONField on `CRUDEvent`, populated when a model defines a hook method.

**Default method name:** `get_easyaudit_metadata(changed_fields)`  
**Configurable via:** `EASY_AUDIT_METADATA_METHOD` on the model

```python
class Order(models.Model):
  account = models.ForeignKey(Account, on_delete=models.CASCADE)

  def get_easyaudit_metadata(self, changed_fields):
      return {"account_id": self.account_id}
```

`changed_fields` is the delta dict for update/M2M events (requires `changed_fields` JSONField PR to be merged first for consistent dict access on updates).

Query related events across models:

```python
CRUDEvent.objects.filter(metadata__account_id=account.id)
```

Models without a metadata method store `metadata=null`.

### 2. Custom audit queryset

When computing update deltas, easy-audit loads the previous row with a model-specific queryset instead of `Model.objects.get()`.

**Default method name:** `get_easyaudit_queryset()`  
**Configurable via:** `EASY_AUDIT_QUERYSET_METHOD` on the model

```python
class MyModel(models.Model):
    @classmethod
    def get_easyaudit_queryset(cls):
        return cls.objects.all_with_deleted()  # e.g. django-safedelete
```

Use case: auditing undelete or other operations where the default manager hides rows.

### 3. Field value resolver map

`get_field_value()` dispatches through `DJANGO_EASY_AUDIT_FIELD_VALUE_RESOLVER_MAP` (settings), keyed by field class.

Default: `DateTimeField` → naive UTC conversion (existing behavior, now pluggable).

```python
# settings.py
DJANGO_EASY_AUDIT_FIELD_VALUE_RESOLVER_MAP = {
    MyCustomField: my_custom_resolver,
}
```

### 4. SimpleCookie import

```python
# Before (removed in newer Django)
from django.http.cookie import SimpleCookie

# After
from http.cookies import SimpleCookie
```

Fixes request-event user resolution when session cookies are parsed.

### 5. M2M field name with multi-table inheritance

When a `ManyToManyField` is declared on a **parent** model but the M2M change happens on a **child** instance, the original `get_m2m_field_name()` logic could return `None`, producing useless `changed_fields` like `{null: [pk, ...]}`.

```python
class Place(models.Model):
    tags = models.ManyToManyField("Tag")

class Restaurant(Place):
    pass

restaurant.tags.add(tag)  # instance is Restaurant; M2M is on Place
```

The fix falls back to `instance._meta.many_to_many`, which includes parent-declared relations, and matches on `related_model`.

## Migration impact

Adds one nullable column:

- `easyaudit/migrations/0021_crudevent_metadata.py` — `AddField(metadata=JSONField(null=True, blank=True))`

Existing `CRUDEvent` rows get `metadata=null`. No data transformation required.

Test app migration `0003_metadata_models.py` adds models used only in tests.

## Breaking changes

**None for projects not using the new hooks.**

Optional considerations:

| Area | Notes |
|------|-------|
| **Custom `LOGGING_BACKEND`** | `crud_info` now may include `"metadata": {...}` or `null` |
| **DB column** | New `metadata` column after migration; raw SQL / reports should account for it |
| **Resolver map** | Replacing `DJANGO_EASY_AUDIT_FIELD_VALUE_RESOLVER_MAP` entirely overrides the default `DateTimeField` resolver unless you re-register it |

## Dependencies

Merge order:

1. **#340** — Django upgrades, async middleware, M2M/UUID handling
2. **changed_fields JSONField PR** — native dict `changed_fields` on update events (metadata hooks receive `changed_fields` as a dict)
3. **This PR**

Branch: `upstream/metadata-extensions` (based on `upstream/changed-fields-jsonfield`).

## Test plan

- [ ] `poetry run pytest tests/test_main.py`
- [ ] `TestMetadataModels::test_metadata_is_populated` — metadata on create/update, custom method name, cross-model query
- [ ] Update a model without metadata hook — `metadata` is `null`
- [ ] Model with `get_easyaudit_queryset` — update delta uses custom queryset (manual or project-specific test)
- [ ] Request event logging with session cookie — no import error on Django 6
- [ ] M2M change on child model with parent-declared M2M — `changed_fields` uses correct field name (e.g. `"tags"`)
- [ ] Run `0021_crudevent_metadata` migration on staging; confirm existing rows unaffected
- [ ] README examples match implemented API


jheld:
Is there any benefit to splitting up conceptual chunks of the work to separate PRs? I no longer have any active projects using this library so bigger reviews are harder for me to fully read through and verify. I do appreciate the write up that mentions which commits the work apply to - I may try to look at those as well in the meantime.

pcraciunoiu:
@jheld I'm open to a split, but would appreciate your thoughts on how you'd like that to be split. 

The changes are not as large as it looks, quite a bit from the django upgrade/lint/docs. So, I could split that off first, and leave the rest here maybe, just have 2 PRs?

